PR #934, opened 2019-04-04
Add status panel for asynch tasks

### LINKED ISSUE #912: Add status panel for asynch tasks
Tasks like geocoding, and to a lesser degree MI ranking, can run for tens of seconds.  They are executed in the background by the server, and the user can continue to interact with the tool, but when they complete  they tend to just update the data without warning/explanation.  Some type of a simple status indicator to indicate what processing tasks are running would help clarify changes that occur on-screen when processing completes.

jryu01:
Resolves #912

In Create Model tab,

- Added a status sidebar with icons that shows the status of the async tasks (currently for variable rankings and geocoding)
- Added a status panel that displays information about the background async requests and also ask user whether to apply the change when tasks are done.


public/store/dataset/actions.ts
```diff
@@ -102,31 +102,44 @@ export const actions = {
...
-				]);
+			mutations.updatePendingRequests(context, { ...update, status: DatasetPendingRequestStatus.RESOLVED });
 			})
 			.catch(error => {
+				mutations.updatePendingRequests(context, { ...update, status: DatasetPendingRequestStatus.ERROR });
 				console.error(error);
 			});
 	},
 
+	fetchDatasetAndVariables(context: DatasetContext, args: { dataset: string, field: string }) {
```
kbirk: Maybe rename this more specifically for the geocoding refresh, since it assumes the field is a geocoded variable.

PR #943, opened 2019-04-22
Cache view data

### LINKED ISSUE #937: MI computation/application should cache
Right now we re-run mutual information scoring whenever we go to the model create or model result screens, regardless of whether or not the target has changed.  We should avoid re-running the scoring in those cases as the result will be unchanged.  It also looks like we need to re-apply the sorting state (driven by the new asynch event panel) each time we navigate back to the model create screen.  We should save that state in the route so that the user doesn't have to re-apply everytime they jump between model result and model create screens.  That goes for anything else that we apply through the asynch panel.

@kbirk can provide guidance on how to best persist the state in the route if you're unsure.

jryu01:
Resolves #937

- Fixed that important badges on the facets are not kept and rendered when create model tab is revisited
- Avoid re-fetching unchanged data when navigating to different view

### DIFF
--- a/public/store/view/actions.ts
+++ b/public/store/view/actions.ts
@@ -17,26 +17,26 @@ function createCacheable(cacheKey: string, func: (context: ViewContext, args: {[
 	};
 }
 
-const updateVariables = createCacheable('fetchVariables', (context, args) => {
+const fetchVariables = createCacheable('variables', (context, args) => {
 	return context.dispatch('fetchVariables', args);
 });
 
-const updateVariableSummaries = createCacheable('fetchSummaries', (context, args) => {
-	return updateVariables(context, args).then(() => {
+const fetchVariableSummaries = createCacheable('variableSummaries', (context, args) => {
+	return fetchVariables(context, args).then(() => {
 		const dataset = args.dataset;
 		const variables = context.getters.getVariables;
 		context.dispatch('fetchVariableSummaries', { dataset, variables });
 	});
 });
 
-const updateVariableRankings = createCacheable('fetchVariableRankings', (context, args) => {
+const fetchVaraibleRankings = createCacheable('variableRankings', (context, args) => {
 	// if target or dataset has changed, clear previous rankings before re-fetch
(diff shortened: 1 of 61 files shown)


public/store/view/actions.ts
```diff
@@ -2,6 +2,50 @@ import _ from 'lodash';
...
+
+const fetchVariableSummaries = createCacheable(ParamCacheKey.VARIABLE_SUMMARIES, (context, args) => {
+	return fetchVariables(context, args).then(() => {
+		const dataset = args.dataset;
+		const variables = context.getters.getVariables;
+		context.dispatch('fetchVariableSummaries', { dataset, variables });
+	});
+});
+
+const fetchVaraibleRankings = createCacheable(ParamCacheKey.VARIABLE_RANKINGS, (context, args) => {
```
kbirk: Typo in method name :)

public/store/view/index.ts
```diff
@@ -1,7 +1,9 @@
 export const LAST_STATE = '__LAST_STATE__';
 
 export interface ViewState {
+	fetchParamsCache: { [key: string]: string };
```
kbirk: You could use the `Dictionary<string>` type for this that can be imported from `util/dict.ts`.

public/store/view/actions.ts
```diff
@@ -2,6 +2,50 @@ import _ from 'lodash';
...
+import { mutations } from './module';
+
+enum ParamCacheKey {
+	VARIABLES = 'VARIABLES',
+	VARIABLE_SUMMARIES = 'VARIABLE_SUMMARIES',
+	VARIABLE_RANKINGS = 'VARIABLE_RANKINGS',
+	SOLUTION_REQUESTS = 'SOLUTION_REQUESTS',
+}
+
+function createCacheable(key: ParamCacheKey, func: (context: ViewContext, args: {[key: string]: string}) => any) {
```
kbirk: `args` Could be `Dictionary<string>` from `util/dict.ts`

PR #1035, opened 2019-06-03
Add condition for close button

### LINKED ISSUE #1012: Geo selection should clear by clicking close button on the selection rectangle
After you drag select something on the geo view and click the X button on the selection rectangle, I would expect the selection to be canceled but neither `X` button disappears nor the selection clears. Instead, it makes 1x1 selection. 

Screen Shot 2019-05-23 at 2 49 41 PM

GAO8A:
- added separate condition for mouse-down event handling for close button
- removed event handler that appears to be unused.

Fixes #1012

### DIFF
--- a/public/components/GeoPlot.vue
+++ b/public/components/GeoPlot.vue
@@ -230,7 +230,7 @@ export default Vue.extend({
 			const mapEventTarget = event.target as HTMLElement;
 			
 			//check if mapEventTarget is the close button or icon
-			if (mapEventTarget.classList.contains(CLOSE_BUTTON_CLASS) ||  mapEventTarget.classList.contains('fa')) {
+			if (mapEventTarget.classList.contains(CLOSE_BUTTON_CLASS) ||  mapEventTarget.classList.contains('fa-times')) {
 				this.removeSelection();
 				return;
 			}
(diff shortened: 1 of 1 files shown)


GAO8A:
It appears the cause of the issue stems from the `onMouseUp` event handler that creates a new selection, regardless if you clicked the close button. The previous event handler (which was dynamically added on creation of the close button) was triggered by a click and worked in non-selection mode. However, the event handlers in the `v-on:` directives seems to override that.

My solution therefore was to check the event target on the `onMouseDown` to see if its the close button or its icon (event on icon doesn't seem to propagate up) and fire the same actions per the escape button event handler.

jryu01:
Ah, and could you include little change to your pr as well?
I've noticed that clicking selction mode button clears the entire url. I think it's because of `href="#"` attributed on line 16
https://github.com/uncharted-distil/distil/blob/2c5affe08544f3b76d1968e2c951dce8cb0473be/public/components/GeoPlot.vue#L16

You could remove href attribute and use css style to make the mouse cursor to pointer or 
use `@click.prevent="isSelectionMode = !isSelectionMode"` directive on the parent elements to prevent default click behaviour

Issue #1027, opened 2019-05-31
Feature appears to render more times with each type selection


karlolsonuc:
As a user toggles between different types for a given feature, it would appear from console logging that each change doesn't just update information once per change, but _n_-times per change where _n_ is the number of previous changes + initial render. 

To view it in action, you'll want to add 

`console.log(currentNormalizedType, nonSchemaSuggestions, missingSuggestions);` 

inside the addMissingSuggestions function of TypeChangeMenu.vue. Then you should see data being set _n_ times before settling in with each change in the browser console, and you can see that the data being set seems to be a like history of the previous changes. This suggests something weird about the state, so further investigation of the underlying cause is needed.

karlolsonuc:
Resolved with my Pull Request that @kbirk merged back.

PR #1039, opened 2019-06-05
Facet Menu Clean Up Fix (#1027)

karlolsonuc:
Since we jquery inject menus into facets, we have to handle the destruction of those elements, otherwise every time we update a facet, we inject a new instance of TypeChangeMenu while the old instances aren't cleaned up (even though they're no longer displayed, their instances are still trying to act on prop changes; they are basically orphaned by the underlying uncharted-software/facets component as it changes, which is why it needs to re-inject every time anyway.) To handle this, its now tracking the menus by their column name as they're created in facets.vue such that if it's already created a menu for that column name, it calls destroy on the old menu before binding a new menu to the facet. It also calls destroy on all current menus when facets.vue is destroyed since the destroy call on our the imported facets instance won't call destroy on the injected menus otherwise, which was resulting in continued attempts at acting on updates even on pages which had no facets on them.

### DIFF
--- a/public/components/Facets.vue
+++ b/public/components/Facets.vue
@@ -1041,9 +1041,7 @@ export default Vue.extend({
 
 		// specifically destroy menus so because we injected them
 		// and so we have to take manual action to destroy them
-		_.forIn(this.menus, function(menu){
-			menu.$destroy();
-		})
+		_.forIn(this.menus, menu => menu.$destroy());
 		this.menus = null;
 	},
 });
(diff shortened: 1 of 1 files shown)


kbirk:
Looks good. One thing to note is that colNames are not uniquely identifying, for example, in the join view the facets will include variables from two datasets and there are cases where two facet groups have the same colName. I would recommend using `${datasetName}:${colName}` instead.

karlolsonuc:
@kbirk made the changes you suggested, everything still seems to work as desired.

PR #1040, opened 2019-06-10
add minutes, hours, and years to sub 4 month category

### LINKED ISSUE #1024: Add finer grained binning intervals
UI currently supports Monthly and Weekly binning intervals.  We should support years, days, hours, minutes and should only display those that make sense for the span and frequency of the selected time data.

GAO8A:
fix #1024


GAO8A:
Deemed unnecessary after reviewing existing binning of time intervals

PR #1049, opened 2019-06-14
Datamart join

### LINKED ISSUE #1007: Display suggested joins
For the summer eval we are supposed to have more robust data mart capabilities.  Specifically, we should be able to issue a join using a dataset as the search criteria, and get back a scored set of joinable datasets, each with a list of columns that are suitable join targets.  We should display these suggested join targets in the top table in green after import, and when a user selects one it should highlight in blue, and then highlight the suggested join targets in green as it currently does.  We can get some input from David about the proper interactions, but this will get us started.

jryu01:
- Integrated with recent datamart api change (search result format has been changed)
- Updated join suggestions route so that join suggestions can be retrieved from datamart without providing search terms but only with dataset id. (Currently only nyu datamart supports the join search)
- Updated join UI so that only suggested columns are highlighted instead of highlighting all joinable columns
- Now interaction in the top table triggers highlights for suggestions/selections/errors in the bottom tables (previously it was bi-directional) 

Resolves #1007 

**Note:**
Currently nyu data mart returns the results as something like this (very simplified).
```
{ id: 1, score: 1,  joinSuggestion: { leftCol: [A], rightCol: [B]}},
{ id: 1, score: 0.1, joinSuggestion: { leftCol: [D], rightCol: [E]}}
```
As a result, user would see two same datasets with different score in the join suggestion panel. And once navigated to the join view, you would generally see only one pair of suggested columns (A, B if you choose first result).
Depending on the design/use case, I think it could be a better option to  merge those two result as a single item in the backend and display all suggested columns in the join view. 

I also noticed that there was a datamart join suggestion result (just one out of many) something like this
`leftCol: "latitude, latitude"` 
`rightCol: "lat, long"`
Implying there could be a join on multiple columns 
I wan't sure if we already support this type of join but I guess it's worth to note.


api/routes/join_suggestion.go
```diff
@@ -88,45 +98,92 @@ func JoinSuggestionHandler(esCtor model.MetadataStorageCtor, metaCtors map[strin
...
 		}
 
-		var deconflicted []*model.Dataset
-		for _, dataset := range exists {
-			deconflicted = append(deconflicted, dataset)
-		}
+		// sort by join score and name
+		sort.Slice(datasets, func(i, j int) bool {
+			return datasets[i].JoinScore > datasets[j].JoinScore || datasets[i].Name < datasets[j].Name
+		})
```
phorne-uncharted: Im not sure this sort works as if join score of one dataset is > then the other but also > for the name then true is returned for both OR clauses. A quick test on https://play.golang.org/p/WG88-O--VD- can illustrate it a bit.

jryu01: I think you are correct. Thanks for catching this. I'll fix!

PR #1176, opened 2019-08-20
Rank Ribbon Display Fixes #1076

### LINKED ISSUE #1076: Rank ribbons only show up after feature pagination or add/remove
When feature ranking results are applied via the asynch panel the features are immediately re-ordered, but the ribbons used to mark highly informative features only display after you add/remove a feature, or if you page the displayed feature set.

karlolsonuc:
Since the rank ribbon now appear to reliably display after applying the ranking (IE: no repro,) just in the wrong place, tighter jquery placement + improved CSS solve our ribbon woes.

### DIFF
--- a/public/components/FacetEntry.vue
+++ b/public/components/FacetEntry.vue
@@ -248,6 +248,12 @@ export default Vue.extend({
 			deep: true
 		},
 
+		ranking(currRanking: number, prevRanking: number) {
+			if (currRanking !== prevRanking) {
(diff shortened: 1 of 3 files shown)


karlolsonuc:
Long story short on the additional changes: jquery injection into the facets + vue store look ups at the single facet level + v-for with ids that weren't always changing resulted in ranking updates having no way to trigger an update at the facet level. By pulling up the ranking look up, making it a prop on the facet, and watching that prop on the facet, updates actually happen.

PR #1187, opened 2019-08-29
Ensures no forecast visuals are shown for timeseries classification

cdbethune:
The table view visuals for timeseries classification problems were displaying a constant forecast value of zero over the ground truth timeseries visual.  This should only be displayed for forecasting problems, so a check has been added.


GAO8A:
LGTM

Issue #1171, opened 2019-08-19
Geocoordinate facet improvements


GAO8A:
- Add type change menu to facet
- Add Mock explode for sub-facets
- Fix facet appearing on dataset change

cdbethune:
A few other immediate items:

- Map should be contained in a card with a title and an icon, similar to other facets
- If we support pan/zoom then we should be recomputing the heatmap after each viewpoint change
- For ACLED dataset, heatmap data doesn't match map view data - there should be hits in Saudi Arabia/Yemen

PR #1193, opened 2019-08-30
Geo facet query

GAO8A:
@phorne-uncharted Please add if neccessary

This PR addresses points in #1171

Note that after creating a geocoordinate facet with the ACLED dataset, a filter param is generated in the URL on initial creation, and on pan/zoom actions on the map. If the page is refreshed, the page cannot return to its previous state as nothing handles the filter params yet.

To temporarily fix this, remove everything after `availableTargetVarsPage=3` in the URL.

Eg.
`host:8080/#/select?dataset=acled_clean&availableTargetVarsPage=3&filters=<really-long-hash>`

becomes 

`host:8080/#/select?dataset=acled_clean&availableTargetVarsPage=3`

Then refresh


api/model/storage/postgres/coordinate.go
```diff
@@ -0,0 +1,331 @@
...
+	}, nil
+}
+
+// FetchPredictedSummaryData pulls predicted data from the result table and builds
+// the coordinate histogram for the field.
+func (f *CoordinateField) FetchPredictedSummaryData(resultURI string, datasetResult string, filterParams *api.FilterParams, extrema *api.Extrema) (*api.VariableSummary, error) {
+	return nil, fmt.Errorf("not implemented")
+}
+
+func (f *CoordinateField) fetchPredictedSummaryData(resultURI string, datasetResult string, filterParams *api.FilterParams, extrema *api.Extrema) (*api.Histogram, error) {
```
golangcibot: U1000: func `(*CoordinateField).fetchPredictedSummaryData` is unused (from `unused`)

public/components/GeocoordinateFacet.vue
```diff
@@ -206,10 +225,28 @@ export default Vue.extend({
 		},
 		rowSelection(): RowSelection {
 			return routeGetters.getDecodedRowSelection(this.$store);
+		},
+		availableTargetVarsPage(): number {
```
cdbethune: Why is this needed?

GAO8A: It's not, good catch, thanks!

Issue #1059, opened 2019-06-20
Error while selecting a histogram of Date/Time facets


jryu01:
When you perform selection on the histogram facets with date/time type, following error occurs. (with acled dataset)

Before:
![Screen Shot 2019-06-20 at 5 31 24 PM](https://user-images.githubusercontent.com/3279838/59883115-075bd080-9382-11e9-811e-9f80ea84f5b3.png)

After:
![Screen Shot 2019-06-20 at 5 35 00 PM](https://user-images.githubusercontent.com/3279838/59883132-1773b000-9382-11e9-9a34-e11107a1d5f4.png)

From server logs: 
[ Jun 20 17:34:33 ] [ ERROR ] code 500:no `min` provided for filter

It looks like following line also return NaN for string date "yyyy/mm/dd" 
https://github.com/uncharted-distil/distil/blob/4e080c59c563bd778dad9007139a7c4e4cd388e6/public/components/FacetEntry.vue#L317

PR #1226, opened 2019-09-19
Date filtering min

### LINKED ISSUE #1060: Date/Time facets doesn't receive row selection highlight
When rows are selected from the table, histogram facets with date/time type doesn't get red highlights.
It looks like it's related to #1059. String date value is being converted to number and returns NaN.
 
https://github.com/uncharted-distil/distil/blob/4e080c59c563bd778dad9007139a7c4e4cd388e6/public/components/FacetEntry.vue#L485

phorne-uncharted:
Fixes #1060 and #1059 

Datetime facets were not handling the string nature of the labels properly. The queries were also not using the proper extrema. Removed the rounding used by other data types when querying for summaries, and added a new datetime filter type that is used when generating where clauses.

### DIFF
--- a/public/components/FacetEntry.vue
+++ b/public/components/FacetEntry.vue
@@ -12,7 +12,7 @@ import Vue from 'vue';
 import IconFork from './icons/IconFork';
 import IconBookmark from './icons/IconBookmark';
 import { createIcon } from '../util/icon';
-import { CATEGORICAL_FILTER, NUMERICAL_FILTER, BIVARIATE_FILTER, FEATURE_FILTER, TIMESERIES_FILTER, INCLUDE_FILTER } from '../util/filters';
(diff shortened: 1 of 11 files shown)


public/components/FacetEntry.vue
```diff
@@ -77,9 +79,11 @@ export default Vue.extend({
 		});
 
 		this.facets.on('facet-histogram:rangechangeduser', (event: Event, key: string, value: any, facet: any) => {
+			const isNumber = !_.isNaN(_.toNumber(value.from.label[0]));
 			const range = {
-				from: _.toNumber(value.from.label[0]),
-				to: _.toNumber(value.to.label[0])
+				from: isNumber ? _.toNumber(value.from.label[0]) : Date.parse(value.from.label[0]) / 1000,
```
cdbethune: I would suggest extracting the range creation out into a function since the logic is repeated in a few places.  Also make the `1000` a constant at the top of the file.

cdbethune: For defining constants see https://github.com/vuejs/vue/issues/6004#issuecomment-312275817

PR #1227, opened 2019-09-20
Adds stratified train/test split function for categoricals

### LINKED ISSUE #931: Support stratified sampling
We currently randomly sample data when we creating our internal train/test split.  It would be better to support running in a stratified sampling mode that groups on target to ensure we get good coverage in cases where the targets are not well balanced.  There may be a TA1 primitive we could execute to do this, or we could implement directly in the server ourselves.

cdbethune:
Fixes #931.  A follow on PR will address re-using splits when parameters match as an optimization.


api/compute/dataset_persist.go
```diff
@@ -264,31 +264,84 @@ func splitTrainTest(sourceFile string, trainFile string, testFile string, hasHea
...
+			key := row[targetCol]
+			if _, ok := categoryRowData[key]; !ok {
+				categoryRowData[key] = [][]string{}
+			}
+			categoryRowData[key] = append(categoryRowData[key], row)
+		}
+
+		// second pass - randomly sample each category to generate train/test split
+		for _, data := range categoryRowData {
+			maxCategoryRows := len(rowData) / len(data) * maxTrainingCount
```
phorne-uncharted: I think this division is reversed as it is currently going to set max cat rows to a multiple of max training count

api/compute/dataset_persist.go
```diff
@@ -299,41 +352,30 @@ func splitTrainTest(sourceFile string, trainFile string, testFile string, hasHea
...
-		return trainSchemaFile, testSchemaFile, nil
-	}
+	// Disabling for now - check should also take into account whether or not stratified samplling was applied.  Would
+	// require persisting information.
+	//
+	// log.Infof("checking folders `%s` & `%s` to see if the dataset has been previously split", trainFolder, testFolder)
+	// if fileExists(trainSchemaFile) && fileExists(testSchemaFile) {
+	// 	log.Infof("dataset '%s' already split", datasetName)
+	// 	return trainSchemaFile, testSchemaFile, nil
+	// }
```
phorne-uncharted: I worry that disabling this check will cause issues from a user perspective if they are trying to compare some results across multiple solution creations.

api/compute/dataset_persist_test.go
```diff
@@ -0,0 +1,122 @@
...
+	assert.Equal(t, 2, categoricalValues["a"])
+	assert.Equal(t, 1, categoricalValues["b"])
+	assert.Equal(t, 1, categoricalValues["c"])
+	assert.Equal(t, 1, categoricalValues["d"])
+}
+
+func TestCacheHit(t *testing.T) {
+	assert.NoError(t, removeTestFiles())
+	params := createTestParams(true)
+	keyMatch, err := checkAndUpdateCacheKey(params)
```
golangcibot: ineffectual assignment to `keyMatch` (from `ineffassign`)

api/compute/dataset_persist_test.go
```diff
@@ -0,0 +1,122 @@
...
+	params = createTestParams(true)
+	keyMatch, err = checkAndUpdateCacheKey(params)
+	assert.NoError(t, err)
+	assert.True(t, keyMatch)
+}
+
+func TestCacheMiss(t *testing.T) {
+	assert.NoError(t, removeTestFiles())
+	params := createTestParams(true)
+	keyMatch, err := checkAndUpdateCacheKey(params)
```
golangcibot: ineffectual assignment to `keyMatch` (from `ineffassign`)

api/compute/dataset_persist.go
cdbethune: Fixed.

api/compute/dataset_persist_test.go
```diff
@@ -0,0 +1,122 @@
...
+	params = createTestParams(true)
+	keyMatch, err := checkAndUpdateCacheKey(params)
+	assert.NoError(t, err)
+	assert.True(t, keyMatch)
+}
+
+func TestCacheMiss(t *testing.T) {
+	assert.NoError(t, removeTestFiles())
+	params := createTestParams(true)
+	_, err := checkAndUpdateCacheKey(params)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

api/compute/dataset_persist.go
cdbethune: Added the caching back in - now takes the split params into account to ensure that re-split occurs when those parameters are modified.

PR #1246, opened 2019-10-05
WIP: 1214 geofacet filter support

GAO8A:
Basic filtering via range selection and table selection. 

Note: Filtering with row selection followed by area selection does not currently work as expect. Possible backend issue is being looked at.


public/util/filters.ts
```diff
@@ -199,6 +199,8 @@ function dedupeRowFilters(filters: Filter[]): Filter[] {
 function addFilter(filters: string, filter: Filter): string {
 	const decoded = decodeFilters(filters);
 	decoded.push(filter);
+	console.log('added filter', filter);
```
karlolsonuc: cut this later of course once the filter issues are sorted

cdbethune:
I'm going to merge this now is as its mostly functional.  I'll log some new issues for things that still need tweaking.

PR #1251, opened 2019-10-11
1213 highlight support

GAO8A:
Added selection logic to geocoordinate map


cdbethune:
The lat/lon data is now being added to the dataset when a geocoordinate feature is present (#1254).  This should allow for the highlight to be calculated using a constant time test against the buckets.

PR #1287, opened 2019-10-29
Support shap values

### LINKED ISSUE #1262: Store returned per-observation shap values
SHAP values that are returned need to be stored so that they can be forwarded to the client.

phorne-uncharted:
Fixes #1262 

Added support for feature weights per row. The weights are returned in the FilteredData type as well as in the overloaded ranking route. If the ranking route specifies a result id and a d3m index then the feature weights for that row are returned.

Since all feature weights need to be stored, every dataset now gets an explain table that mirrors the dataset structure, with every feature being a double. When fields are added, the table needs to be updated to reflect the new field.

### DIFF
--- a/api/compute/explain.go
+++ b/api/compute/explain.go
@@ -81,7 +81,10 @@ func (s *SolutionRequest) parseSolutionFeatureWeight(resultURI string, outputURI
 		return nil, errors.Wrap(err, "unable to read feature weight output")
 	}
 
-	setD3MIndex(0, d3mIndexLookup, res)
+	err = setD3MIndex(0, d3mIndexLookup, res)
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to update d3m index")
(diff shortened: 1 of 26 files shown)


api/compute/explain.go
```diff
@@ -48,56 +60,34 @@ func (s *SolutionRequest) explainOutput(client *compute.Client, solutionID strin
...
-			return nil, errors.Wrap(err, "unable to parse feature weight")
-		}
-		featureName := res[0][i].(string)
-		weights[i] = &api.SolutionFeatureWeight{
-			SolutionID:  solutionID,
-			FeatureName: featureName,
-			Weight:      weight,
-		}
-	}
+	setD3MIndex(0, d3mIndexLookup, res)
```
golangcibot: Error return value of `setD3MIndex` is not checked (from `errcheck`)

PR #1313, opened 2019-11-09
Add grouping compose

### LINKED ISSUE #1237: Properly integrate SuggestedGroupingKey and GroupingKey types
Grouping timeseries data relies on two new(ish) semantic types: `SuggestedGroupingKey` and `GroupingKey`.  The idea is that a provided dataset will have a `SuggestedGroupingKey` tagged against columns that are potential key components that TA2s can leverage to infer a  final key on their own.  The final key should be contained in a new column and marked with type `GroupingKey` - this is what TA1 primitives should use to split the timeseries that are stored in long form.  TA3s can use the `SuggestedGroupingKey` as hint, or ignore it it and just let the user actions determine the splitting key construction - as long as there is a column marked with `GroupingKey` that makes its way to the consuming TA1 primtives everything should be good.

In order for all of this to work properly, our TA3 system should be creating a new column in the DB for the composed key (we do), adding a key composition primitive (doesn't exist yet) to our preprocessing pipeline to handle combining `species` and `sector` into a single key at pipeline run time, and marking the produced column with a the `GroupingKey` semantic type so that TA1 primitives can properly handle them.

phorne-uncharted:
Fixes #1237 

Update to use target feature rather than target name when building pre processing.


api/compute/solution_request.go
```diff
@@ -595,19 +611,16 @@ func (s *SolutionRequest) PersistAndDispatch(client *compute.Client, solutionSto
...
-		if variable.Name == s.TargetFeature {
-			targetVariable = variable
-		}
 	}
+	targetVariable := s.TargetFeature
 
 	// Timeseries are grouped entries and we want to use the Y Col from the group as the target
 	// rather than the group itself, and the X col as the timestamp variable
-	targetVarName := s.TargetFeature
+	targetVarName := targetVariable.Name
```
golangcibot: ineffectual assignment to `targetVarName` (from `ineffassign`)

PR #1350, opened 2019-12-03
Restrict explain shap

### LINKED ISSUE #1249: Optimize feature weight pipeline
The ensemble forest primitive has 2 arguments to control the number of forests it uses:
- `small_dataset_fits`
- `large_dataset_fits`
To speed up execution of the feature weights pipeline, we should force both to `1` prior to running it.  Additionally, there is the `dataset_sample` common primitive that allows for training rows to be sampled.  Inserting that primitive and pushing the row sample count down to a few hundred entries would significantly speed up execution (at the cost of weight accuracy).

phorne-uncharted:
Fixes #1249

Switched the explain workflow to search on an initial solution, describe the resulting pipeline, then modify that pipeline to also output the shap values. In doing so, the fit & run are only done once.


api/compute/solution_request.go
```diff
@@ -342,7 +343,7 @@ func (s *SolutionRequest) createProduceSolutionRequest(datasetURI string, fitted
 func (s *SolutionRequest) persistSolutionError(statusChan chan SolutionStatus, solutionStorage api.SolutionStorage, searchID string, solutionID string, err error) {
 	// persist the updated state
 	// NOTE: ignoring error
-	solutionStorage.PersistSolution(searchID, solutionID, SolutionErroredStatus, time.Now())
+	solutionStorage.PersistSolutionState(solutionID, SolutionErroredStatus, time.Now())
```
golangcibot: Error return value of `solutionStorage.PersistSolutionState` is not checked (from `errcheck`)

PR #1375, opened 2019-12-17
Create produce route

phorne-uncharted:
Added a new route that calls produce on a fitted pipeline. The new data is ingested into the database, then the prediction results are also ingested.


api/task/ingest.go
```diff
@@ -290,6 +297,7 @@ func Ingest(originalSchemaFile string, schemaFile string, storage api.MetadataSt
 	// Hardcoded the base table name for now.
 	pg.DropView(dbTable)
 	pg.DropTable(fmt.Sprintf("%s%s", dbTable, baseTableSuffix))
+	pg.DropTable(fmt.Sprintf("%s%s", dbTable, explainTableSuffix))
```
golangcibot: Error return value of `pg.DropTable` is not checked (from `errcheck`)

PR #1376, opened 2019-12-17
Timeseries pattern cleanup

cdbethune:
1.  Calls cluster route when time series is created
1.  Adds logic to replace timeseries facet entries with pattern facet entries when clusters are available
1.  Ensures all facet highlight operations support the dual-mode timeseries facet display

There is still some additional work to be done client side to ensure that:
1. The time series cluster variable does not appear as a facet (should only appear in timeseries facet)
1. The display of the cluster data in the timeseries facet only occurs if the user applies the cluster info from the asynch panel
That work will be done in a follow on PR.

### DIFF
--- a/api/model/storage/postgres/field.go
+++ b/api/model/storage/postgres/field.go
@@ -17,6 +17,7 @@ package postgres
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/pkg/errors"
 	"github.com/uncharted-distil/distil-compute/model"
@@ -28,7 +29,8 @@ type Field interface {
 	FetchSummaryData(resultURI string, filterParams *api.FilterParams, extrema *api.Extrema, invert bool) (*api.VariableSummary, error)
 	FetchPredictedSummaryData(resultURI string, datasetResult string, filterParams *api.FilterParams, extrema *api.Extrema) (*api.VariableSummary, error)
 	GetStorage() *Storage
-	GetStorageName() string
+	GetDatasetStorageName() string
+	GetDatasetName() string
 	GetKey() string
 	GetLabel() string
 	GetType() string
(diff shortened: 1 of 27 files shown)


api/model/storage/postgres/field.go
```diff
@@ -16,11 +16,104 @@
...
+		}
+	}
+	return nil
+}
+func (b *BasicField) hasClusterData(variableName string) bool {
+	query := fmt.Sprintf("SELECT EXISTS (SELECT 1 FROM information_schema.columns WHERE table_name = '%s' AND column_name = '%s');",
+		b.DatasetStorageName, variableName)
+	res, err := b.Storage.client.Query(query)
+	if err != nil {
+		errors.Wrap(err, "failed to query cluster column status")
```
golangcibot: Error return value of `errors.Wrap` is not checked (from `errcheck`)

api/model/storage/postgres/filter.go
```diff
@@ -615,3 +626,21 @@ func (s *Storage) FetchData(dataset string, storageName string, filterParams *ap
...
 	return s.parseFilteredData(dataset, variables, numRows, res)
 }
+
+func (s *Storage) hasClusterData(storageName string, colName string) bool {
+	clusterColName := getClusterColName(colName)
+	query := fmt.Sprintf("SELECT table_name FROM information_schema.columns WHERE table_name = %s AND column_name = %s;",
+		storageName, clusterColName)
+	res, err := s.client.Query(query)
+	if err != nil {
+		errors.Wrap(err, "failed to query cluster column status")
```
golangcibot: Error return value of `errors.Wrap` is not checked (from `errcheck`)

Issue #1368, opened 2019-12-10
Interface for running a model against new data


cdbethune:
For the WM exercise, we need to be able to load a CSV file, and run it against the currently selected model.  As a placeholder, we can just add a button next to "Export" on the results screen called "Infer", which will let a user select a CSV file to act as an additional data source.  Once that is loaded, we need to invoke a new `produce` route, passing it the URI of the CSV file and the model (solution ID).  The route is being developed under #1367, and the new screen to explore the predictions is being developed under #1368.

cdbethune:
@djonker, @slangevi  - is this roughly the direction we want to go in?

cdbethune:
Reviewed with @slangevi.  We will start moving in this direction for the Dec. 20th demo.

PR #1385, opened 2019-12-19
#1368 - inference upload

karlolsonuc:
Various changes that allow the file uploader and supporting actions to be reused, it's use in place of export on model page, additional changes to support new prediction view to be built in the next task, updated store, utilities and dependent components to handle and use fitted solution id to post the prediction data.


public/store/results/actions.ts
```diff
@@ -21,6 +21,7 @@ import {
 } from "../../util/data";
 import { getters as resultGetters } from "../results/module";
 import { getters as dataGetters } from "../dataset/module";
+import { curveCatmullRomClosed } from "d3";
```
cdbethune: Accidentally included?

karlolsonuc: yeah that's definitely weird. auto complete and auto include.

PR #1405, opened 2020-01-07
New var instances for grouped variables

### LINKED ISSUE #1393: Geocoordinate facets report network error after a type change
After a geo coordinate is created, if another a facet has its type changed, the geocoordinate is replaced with `LONGITUDE` and reports a network error in the facet:

image

cdbethune:
Previously creating geocoord and timeseries grouped variables lead to two different paths being followed.  For a geocoord, we would modify the longitude variable so that it acted as a grouped variable, and we would hide the associated longitude.  For timeseries, we would create a new variable to hold a grouping key, and then use that new variable as the grouped timeseries variable.  This lead to a lot of slightly different conditional logic being added to the system to handle both cases, and was extremely fragile and buggy.  

As a first step in a larger variable cleanup, we now create a new grouped variable in both cases, and set its grouping structure to reference its constituent components.  This allows for a lot more consolidation of code, and provides required flexibility when creating timeseries.

Fixes #1393.


api/model/storage/postgres/result.go
```diff
@@ -167,28 +170,17 @@ func (s *Storage) PersistResult(dataset string, storageName string, resultURI st
 		log.Warnf("Result contains %d columns, expected 2.  Additional columns will be ignored.", len(records[0]))
 	}
 
-	targetName := target
+	// Fetch the actual target variable (this can be different than the requested target for grouped variables)
+	targetVariable, err := s.getResultTargetVariable(dataset, target)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

api/routes/variable_rankings.go
```diff
@@ -73,7 +73,12 @@ func VariableRankingHandler(metaCtor api.MetadataStorageCtor, solutionCtor api.S
 			}
 			rankings, err = featureRank(dataset, resultURI, d3mIndex, dataCtor)
 		} else {
-			rankings, err = targetRank(dataset, target, d.Folder, d.Variables, d.Source)
+			summaryVariables, err := api.FetchSummaryVariables(dataset, storage)
+			if err != nil {
+				handleError(w, errors.Wrap(err, "unable to expand grouped variables"))
+				return
+			}
+			rankings, err = targetRank(dataset, target, d.Folder, summaryVariables, d.Source)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

PR #1409, opened 2020-01-10
Allows timeseries with no grouping key selected

### LINKED ISSUE #1174: Support single timeseries forecasting
We currently don't properly support the case where we are forecasting on a dataset that consists of a single timeseries - the most relevant example is the `56_sunspots` dataset:

image

When we use the `Timeseries..` menu entry to create the compound facet, we don't force the user to select as a series ID column, but then when we submit the compound facet isn't created.

cdbethune:
Fixes #1174.  We now have the option of not specifying a grouping key column - in that case the column data is treated as a single time series.

### DIFF
--- a/api/routes/grouping.go
+++ b/api/routes/grouping.go
@@ -71,14 +71,17 @@ func GroupingHandler(dataCtor api.DataStorageCtor, metaCtor api.MetadataStorageC
 		}
 
 		if model.IsTimeSeries(grouping.Type) {
-			// Create a new variable and column for the time series key.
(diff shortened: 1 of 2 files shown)


api/routes/grouping.go
```diff
@@ -200,9 +203,19 @@ func createComposedVariable(metaStorage api.MetadataStorage, dataStorage api.Dat
 		}
 	}
 
-	// Fetch data using the source names as the filter
-	filter := &api.FilterParams{
-		Variables: sourceVarNames,
+	composedData := map[string]string{}
+	var filter *api.FilterParams = nil
```
golangcibot: ineffectual assignment to `filter` (from `ineffassign`)

PR #1462, opened 2020-02-10
Image clustering pipeline

phorne-uncharted:
Added image clustering to distil. Once the type is switched to image, the pipeline is kicked off. The cluster filter type was added to properly generate where clauses on the backend.

### DIFF
--- a/api/task/cluster.go
+++ b/api/task/cluster.go
@@ -24,6 +24,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/uncharted-distil/distil-compute/metadata"
 	"github.com/uncharted-distil/distil-compute/model"
+	"github.com/uncharted-distil/distil-compute/pipeline"
 	"github.com/uncharted-distil/distil-compute/primitive/compute/description"
 	"github.com/uncharted-distil/distil-compute/primitive/compute/result"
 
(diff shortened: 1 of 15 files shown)


api/task/cluster.go
```diff
@@ -124,26 +125,44 @@ func ClusterDataset(datasetSource metadata.DatasetSource, schemaFile string, ind
...
+		}
+	}
+
+	var step *pipeline.PipelineDescription
+	var err error
+	addMeta := false
+	if model.IsImage(clusteringVar.Type) {
+		step, err = description.CreateImageClusteringPipeline("business", "basic image clustering", []*model.Variable{clusteringVar})
+		if err != nil {
+			return false, nil, errors.Wrap(err, "unable to create image clustering pipeline")
```
cdbethune: We probably don't need to call `errors.Wrap` on most of these errors.  Since its our code, we should already be creating the error using `errors.Errorf` further down the call stack.  Adding the `Wrap` just generates more log noise.  It looks like the pre-existing code was wrapping when it shouldn't have been as well.

public/util/types.ts
```diff
@@ -95,9 +95,9 @@ const FLOATING_POINT_TYPES = [
   LONGITUDE_TYPE
 ];
 
-const FEATURE_TYPES = [IMAGE_TYPE];
+const FEATURE_TYPES = [];
```
cdbethune: We can probably just take out all the feature related code.  I'll log an issue.

PR #1471, opened 2020-02-12
Cluster cache fixes and View Models data refresh

### LINKED ISSUE #1458: Clustering and ranking running too often
Clustering and ranking seem to both run far too often.  For example, going from the create screen to the result screen seems to result in both running, even though the data should be unchanged.

cdbethune:
Fixes #1458 

1. Ensures that cluster function is properly integrated into caching mechanism provided by the `view` module
1. Ensures that cluster + ranking cache state is not wiped out by forced refresh of result data
1. Bonus websocket fix - websockets were never being cleaned up, and associated `Stream` objects were kept in the Vue store (which they shouldn't be, because they are a complex object that is not driving UI)

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -11,7 +11,7 @@ fi
 export SOLUTION_COMPUTE_ENDPOINT=localhost:45042
 export ES_ENDPOINT=http://localhost:9200
 export SOLUTION_COMPUTE_TRACE=true
-export PG_LOG_LEVEL=debug # debug, error, warn, info, none
+export PG_LOG_LEVEL=none # debug, error, warn, info, none
 export SKIP_INGEST=true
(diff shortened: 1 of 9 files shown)


run.sh
```diff
@@ -11,7 +11,7 @@ fi
 export SOLUTION_COMPUTE_ENDPOINT=localhost:45042
 export ES_ENDPOINT=http://localhost:9200
 export SOLUTION_COMPUTE_TRACE=true
-export PG_LOG_LEVEL=debug # debug, error, warn, info, none
+export PG_LOG_LEVEL=none # debug, error, warn, info, none
```
karlolsonuc: do we want to leave this on none?

PR #1478, opened 2020-02-14
disables unsupported targets rather than removing them

### LINKED ISSUE #1461: Image name facet not showing in target screen
Image filename variables show up as having a text type initially.  If you change the type to Image in the target screen, the facet will no longer be visible.  The facet does appear in subsequent screens.

cdbethune:
fixes #1461 

We were hiding unsupported targets which ended up looking like a bug because they dissappeared after you updated their type.  We now just disable them.

### DIFF
--- a/public/components/AvailableTargetVariables.vue
+++ b/public/components/AvailableTargetVariables.vue
@@ -18,6 +18,7 @@
 
 <script lang="ts">
 import "jquery";
+import _ from "lodash";
 import {
   getters as datasetGetters,
   actions as datasetActions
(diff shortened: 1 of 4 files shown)


public/components/AvailableTargetVariables.vue
```diff
@@ -18,6 +18,7 @@
 
 <script lang="ts">
 import "jquery";
+import _ from "lodash";
```
karlolsonuc: doesn't look like this is used to me.

PR #1526, opened 2020-04-01
Prediction flow

cdbethune:
Updates to the prediction workflow to resolve issues related to state management.

1. Runs prediction requests / updates through the websocket interface similar to Solutions.
2. Stores prediction state in the server-side DB similar Solutions.
3. Updates PredictionsView to fetch prediction request data from server on mount, ensuring that state is properly restored.
4.  Ensures that prediction data is used for training summaries in prediction screen.
5.  A number of minor tweaks, refactors, clean ups that were low impact.

There is still a quite a bit of code that is common to predictions and solutions that could be factored out, but this change was getting pretty large so it wasn't tackled.  There are some definite rough edges in the UI that still need to be handled, and we still need update he predictions task so that it provides periodic status updates while running, similar to solutions.  All of this should be reasonably straightfoward now that the baseline is in place.


api/compute/predictions_request.go
```diff
@@ -0,0 +1,98 @@
...
+type PredictRequest struct {
+	DatasetID        string
+	Dataset          string
+	TargetType       string
+	FittedSolutionID string
+	TimestampField   string
+	MaxTime          int
+
+	requestChannel chan PredictStatus
+	listener       PredictStatusListener
```
golangcibot, line 21: field `listener` is unused (from `unused`)

Issue #1521, opened 2020-03-30
Discriminate pos/neg SHAP in visuals


cdbethune:
We currently take `abs(shap_value) ` and use that to generate an index into a single blue color rampl.  We are giving up directional information that is useful to the user, and should try to introduce a means by which we incorporate the sign into the visuals.  The semantics of this are not entirely clear, but we can at least see what a naive implementation looks like where we map pos/neg to two distinct colour ramps.  We can use red/blue for now, although red is used to capture errors and would need to be replaced.

An explanation of SHAP values that may help with the semantic interpration can be found here:

https://christophm.github.io/interpretable-ml-book/shapley.html

cdbethune:
Addresed by #1527.

PR #1527, opened 2020-04-01
1279 - shap weight viz

### LINKED ISSUE #1279: Visualize SHAP values
#1278 adds placeholder UI for displaying SHAP values for observations.  SHAP values are weights assigned to each variable in a result row, indicating how that variable contributed to either an  decrease in the predicted value.  

The source code for the SHAP libraries contains some functions to generate visualizations that provide a good visual explanation of the idea:

image

We need to effectively visualize SHAP values within Distil.

karlolsonuc:
Fixes #1279: the initial pass at SHAP viz was based on absolute impact on the score, and ignored the positive/negative or under/over aspect of the value regarding impact on the predicted value from the model. Now, if a SHAP weight is positive, it'll be blue, and if negative, it'll be pink, with a floor on coloration such that it doesn't fade into a hard to pick out off-white as it approaches 0. 0 weight values are still white. Additionally, the autoMpg set included in the repo has been updated to the min-metadata version, and the .gitignore has been updated to explicitly expect that folder to be in the repo so future updates are more easily handled as well.


cdbethune:
Clarification - this was addressing #1521.  #1279 is track outstanding design work.

PR #1563, opened 2020-04-13
Update Bootstrap Vue to Latest

karlolsonuc:
Just to make sure we're pointed at a version with bootstrap date and time pickers as they'll be useful for the upcoming prediction workflow changes. Appears to have resulted in no regressions.


karlolsonuc:
One sec, should do the package files too.

PR #1564, opened 2020-04-13
Show all predictions for a model + other prediction view fixes

### LINKED ISSUE #1556: Show all prediction results for fitted solution
In the predictions view we should display all results associated with a given fitted solution ID, and allow the user to switch between them, similar to the model result view.

cdbethune:
fixes #1556 

- Fixes ingested prediction types so that vars in the prediction screen don't all display in categorical facets
- Displays all predictions made against a selected solution, and allows the user to switch between them similar to the "view models" screen
- Enables highlighting on the prediction summary facets
- Allows proper transitioning back to "view models" from the predictions screen


pg Session.sql
```diff
@@ -0,0 +1,13 @@
...
+-- FROM d_morelearningdata data
+-- INNER JOIN d_morelearningdata_result result ON data."d3mIndex" = result.index
+-- WHERE
+--   result.result_id = '/home/chris/outputs/supporting_files/9dd5f02c-7945-43a4-a89e-b126cf0cf63b_outputs.0.csv'
+--   AND "horsepower" != 'NaN';
+SELECT
+  *
+FROM d_morelearningdata_result data
+LIMIT
+  1000
```
phorne-uncharted, line 13: Is this file necessary???

PR #1587, opened 2020-04-22
Predictions table break out

cdbethune:
Predictions were re-using Results table components wholesale, which was getting a bit messy due to the way data is fetched, and the predictions UI being a simpler case than Results.  Predictions now have their own higher level components, although the fixed table, mosaic components etc. are still common.

### DIFF
--- a/public/store/dataset/actions.ts
+++ b/public/store/dataset/actions.ts
@@ -265,6 +265,8 @@ export const actions = {
       status: DatasetPendingRequestStatus.PENDING
     };
 
+    const sheblag = getters.getVariables(context);
+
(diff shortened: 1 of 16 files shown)


public/store/dataset/actions.ts
```diff
@@ -265,6 +265,8 @@ export const actions = {
       status: DatasetPendingRequestStatus.PENDING
     };
 
+    const sheblag = getters.getVariables(context);
```
karlolsonuc, line 268: This looks like it's unused - I think we can cut it?

cdbethune: Yeah, left over debugging cruft.  I'll get it now.

PR #1597, opened 2020-05-01
Add custom map tile requests #1580

YohannParis:
Test to see if we can display satellite images with Leaflet.


public/components/GeoPlot.vue
```diff
@@ -232,8 +234,32 @@ export default Vue.extend({
     mapZoom(): number {
       return routeGetters.getGeoZoom(this.$store);
     },
+
     rowSelection(): RowSelection {
       return routeGetters.getDecodedRowSelection(this.$store);
+    },
+
+    isSatelliteImages(): boolean {
```
YohannParis, line 242: I create a simple `boolean` to know if the dataset contains satellite images or not.
Maybe this will be done by a primitives?

public/components/GeoPlot.vue
```diff
@@ -232,8 +234,32 @@ export default Vue.extend({
...
+      return datasetGetters.isSatelliteImages(this.$store);
+    },
+
+    currentSatelliteBand(): SatelliteBand {
+      return appGetters.getCurrentSatelliteBand(this.$store);
```
YohannParis, line 247: @cdbethune said that the satellite band will be a "root" variable define by the user, this is why I add it to the `store/app`.

public/components/GeoPlot.vue
```diff
@@ -490,32 +516,41 @@ export default Vue.extend({
...
-            },
-            { animate: true }
-          );
-        }
-        this.baseLayer = leaflet.tileLayer(
-          "http://{s}.basemaps.cartocdn.com/light_all/{z}/{x}/{y}.png"
+    /**
+     * Create a Leaflet map, if it doesn't exist already, with basic defaults.
+     */
+    createMap() {
```
YohannParis, line 522: the `paint()` was big, so I thought of moving out  the creation of the map because it was easy.

public/store/app/index.ts
```diff
@@ -9,13 +10,20 @@ export interface AppState {
   statusPanelState: StatusPanelState;
 }
 
+export interface SatelliteBand {
```
YohannParis, line 13: From what I read, the band are not defined by a mix of RGB colours, but by bands number.
Or is the goal of the satellite band primitive to make false colours by transforming up to 3 bands into RGB?

public/store/app/index.ts
```diff
@@ -9,13 +10,20 @@ export interface AppState {
...
+}
+
 export interface StatusPanelState {
   isOpen: boolean;
   contentType: StatusPanelContentType;
 }
 
 // shared data model
 export const state: AppState = {
+  currentSatelliteBand: { r: 4, g: 3, b: 2 },
```
YohannParis, line 26: I just used the band number for `Landsat 8-9` for RGB. But this might be wrong based on my previous comment on the `interface SatelliteBand`.

public/components/GeoPlot.vue
cdbethune: We're figuring out the best way to do this now.  There are two options: it gets flagged in the `Dataset` data that comes back from the `distil/datasets/<name>` route, so you would be able to pull out of the `store/datasets/` vuex module, or it will be something to check in the returned `Task` data that we fetch via `distil/task/<dataset>/<target>/<var-names>` and store in `store/datasets` as well.  Either way, you won't have to worry about fetching the data from there server, you'll just need to update this function to check the store for one of the cases above.

public/components/GeoPlot.vue
cdbethune: I think we got signals crossed on "root" vs. "route".  What you will have access to in the store is a stucture similar to:

```typescript
interface BandMapping {
    mappingID: string, // needs to be unique
    displayName: string, // exact string to display in UI
    mapping: int[] // length of 3 containing band mapping for r, g, b in order
}
```
and we'll have to store the active `mappingID` in the route (store/route) so that we have it on refresh.

Issue #1578, opened 2020-04-21
Mutliband image support in image views


cdbethune:
We currently have table and image tile views that support the display of standard RGB images.  We should be able to display multiband images in these views as well.  Assuming the addition of the route described in #1576 this should not require heavy processing of imagery on the client side.  Addition of map-based display of imagery will be covered under another feature.

PR #1618, opened 2020-05-11
#1578: multiband image display

karlolsonuc:
- Change to satellite.go to close gdal references.
- Fixes to main.go routes for multiband-image & multiband-combinations
- Added fetchMultiBandImage to dataset/module
- Added fetchMultiBandImage to dataset/actions, also cleaned up arg validation in actions to leverage new validation function added in data.ts util.
- Added multibandimage.go to postgres api so variable.go could have a multibandimage case.
- Fixed a misspelling in imagery.go as I needed to mirror those constants on the client side with new util/bands.ts file.
- Updated the dataset/actions.ts to catch and handle the multiband image case, then extended the inject image preview in the facet entry to handle the multiband image case.
- Added the vue-store related actions needed to retrieve bands, though at this phase of the work, we're not yet making the call at any point. This sets up the next stage of allowing the user to choose the bands combinations they'd like to use for display/analysis.


karlolsonuc:
Okay, I'll tackle those changes and then give you aheads up @cdbethune.

PR #1625, opened 2020-05-15
Remote sensing grouping

phorne-uncharted:
Got rid of duplicated values in for constants from compute.

The biggest difference for handling remote sensing was in the need to updated the train & test splitting. Tiles need to be split together (all bands must end up in the same set). To achieve this goal, the tile is randomly sent to train or test when the first band is encountered, and all subsequent bands for that tile is sent to the same file.


cdbethune:
I'll merge once the conflics are resovled.

PR #1634, opened 2020-05-28
Added a check to make sure that `Remove All` button only applies to non timeseries.

### LINKED ISSUE #1611: Remove all shouldn't remove grouping feature for timeseries
When doing timeseries forecasting, the grouping features (eg. `species` and `sector` for spawn) are automatically included in the features to model and have no remove button as they have to be included in the dataset.  `Remove All` still allows for them to be removed.

YohannParis:
closes #1611 

I'm not sure this solves the problem. I keep looking into the source code to understand how time-series are flagged, but I do not seem to find a ~proper~ way to detect when the predicted feature is a time-series.

### DIFF
--- a/public/components/TrainingVariables.vue
+++ b/public/components/TrainingVariables.vue
@@ -24,7 +24,7 @@
         <div>
           {{ subtitle }}
         </div>
-        <div v-if="trainingVariableSummaries.length > 0">
+        <div v-if="isAllTrainingVariablesRemovable">
           <b-button size="sm" variant="outline-secondary" @click="removeAll"
             >Remove All</b-button
           >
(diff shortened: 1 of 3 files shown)


cdbethune:
I think you should just be able to check the top level type.  I looked at the `variables` entry in the `dataset` vuex module, and the timeseries variable count is defined as:

image

The grouping check you implement is valid as well, but I think implementing the check just on the `colType` field should be sufficient.

YohannParis:
@cdbethune I keep looking how to do it using the `colType` property of the `dataset/variables[]`, but in the example of `Spawn` dataset: `species` and `sector` have their `colType` property set to `categorical` not `timeseries`.

Maybe I misunderstood, but the goal of the bug is when you set a timeseries with certain features, like so:

Screen Shot 2020-06-01 at 16 00 07

Then you should not be able to remove the `Series ID` (`species` and `sector` here) from the `Features to Model` list?

Screen Shot 2020-06-01 at 15 59 29

PR #1635, opened 2020-05-29
1612 features in grouping should not be available for target

### LINKED ISSUE #1612: Features included in compound facets should no longer be visible in the target screen
If we create a timeseries compound facet, the features used as grouping keys should no longer be visible in the target selection screen.

YohannParis:
closes #1612 

Add a filters in the computed `summaries` of `AvailableTargetVariables` component.
This should work if multiple grouping occurs on one dataset.

### DIFF
--- a/public/components/AvailableTargetVariables.vue
+++ b/public/components/AvailableTargetVariables.vue
@@ -59,10 +59,26 @@ export default Vue.extend({
     dataset(): string {
       return routeGetters.getRouteDataset(this.$store);
     },
+
     summaries(): VariableSummary[] {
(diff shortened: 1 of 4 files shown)


YohannParis:
I think I made a mistake, this one also includes the fix for the #1611 issue.
Once #1611 is closed, I will update this pull request.

cdbethune:
This looks good, but the comments from #1634 need to be addressed.

YohannParis:
The comment from #1634 has been addressed, and I merge the latest version of `master` so this is good to go!

PR #1662, opened 2020-06-16
Remote sensing grouping

### LINKED ISSUE #1652: Remove special cases for GEOCOORDINATE_TYPE
Doing a global search on `GEOCOORDINATE_TYPE` reveals a number places where a type specific code path is executed for geo coordinates only.  After a quick scan of the code, it seems likely these branches are no longer necessary - this should be verified and the code removed if true.

### LINKED ISSUE #1658: Remote sensing compound facet
Remote sensing on the client side needs to work as a `remote_sensing` grouped variable, and be integrated with the table, and geocoordinate facet.

cdbethune:
fixes #1658, fixes #1652 

Creates a compound facet for remote sensing data.  This is an interim step, as it creates the facet directly from the type change handler and makes assumptions about the structure of the data.  At minimum, we should be creating the group on ingest/import, or providing a way to build it from existing variables (similar to geo/timeseries).


pg.session.sql
```diff
@@ -0,0 +1,7 @@
+-- SELECT MIN("coordinates" [1]) AS "min_x",
+--     MAX("coordinates" [5]) AS "max_x",
+--     MIN("coordinates" [2]) AS "min_y",
+--     MAX("coordinates" [6]) AS "max_y"
+-- FROM d_bigearth_tiny;
+SELECT "coordinates[1]"
+FROM d_bigearth_tiny;
```
phorne-uncharted, line 7: Pretty sure this file should not be included

PR #1748, opened 2020-07-11
Hdb kmeans config

phorne-uncharted:
Added environment variable to determine if clustering will use kmeans or hdbscan.


cdbethune:
Looks good other than the conflicts.

PR #1793, opened 2020-07-22
Add SHAP legend

### LINKED ISSUE #1777: Add SHAP legend
We should have some type of a legend when we display SHAP values.  One possibility would be a horizontal ramp next to the `Samples` title:

image

YohannParis:
closes #1777

- Created a new component `LegendSHAP`.
- Used the same method to show legend colour has the one used for to colour the cell on the result table of `ResultsComparison` component.

### DIFF
--- a/public/components/ResultsComparison.vue
+++ b/public/components/ResultsComparison.vue
@@ -8,7 +8,8 @@
       :variables="variables"
       class="view-toggle"
     >
-      <p class="font-weight-bold mr-auto">Samples</p>
+      <p class="font-weight-bold">Samples</p>
+      <legend-shap v-if="isTableView" class="ml-5 mr-auto" />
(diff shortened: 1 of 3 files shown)


public/components/ResultsComparison.vue
```diff
@@ -8,7 +8,8 @@
       :variables="variables"
       class="view-toggle"
     >
-      <p class="font-weight-bold mr-auto">Samples</p>
+      <p class="font-weight-bold">Samples</p>
+      <legend-shap v-if="isTableView" class="ml-5 mr-auto" />
```
YohannParis, line 12: I don't know if this `isTableView` is necessary. I don't know if we are displaying SHAP on result other than TABLE_VIEW.

PR #1803, opened 2020-07-24
Update the weight legend.

### LINKED ISSUE #1777: Add SHAP legend

YohannParis:
closes #1777

- update the name of the component to be `LegendWeight` instead of `LegendSHAP`.
- Added a test to display the legend if one of the result value contains a `weight` property.

### DIFF
--- a/public/components/ResultsComparison.vue
+++ b/public/components/ResultsComparison.vue
@@ -8,8 +8,8 @@
       :variables="variables"
       class="view-toggle"
     >
-      <p class="font-weight-bold">Samples</p>
-      <legend-shap v-if="isTableView" class="ml-5 mr-auto" />
+      <p class="font-weight-bold" :class="{ 'mr-auto': !hasWeight }">Samples</p>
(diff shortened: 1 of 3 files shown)


public/components/ResultsComparison.vue
```diff
@@ -109,8 +109,12 @@ export default Vue.extend({
       return this.viewTypeModel;
     },
 
-    isTableView(): boolean {
-      return this.viewType === TABLE_VIEW;
+    /* Check if any items display on the table have a weight property. */
+    hasWeight(): boolean {
+      const data = this.includedTableDataItems ?? [];
+      return data.some(item =>
+        Object.values(item).some(variable => variable.hasOwnProperty("weight"))
```
cdbethune, line 116: Just for reference, lodash has a nice impl of this:  https://lodash.com/docs/4.17.15#get

PR #1821, opened 2020-07-27
Forecasting workflow

### LINKED ISSUE #1370: Forecasting workflow
We currently rely on the D3M problem specification to provide us with information about the forecast start time and number of intervals, rather than allowing the user to set them.  A model is then trained using that fixed forecast period and the ground truth data, and displayed to the user in the result screen.  We need to develop a workflow that allows the user to set the forecasting horizon themselves in some capacity.   

This could be done on the create screen, and the result screen would allow the user to view the forecasted values, along with error information you can calculate by pushing your entire data period through the model.  No explicit "ineference" step would be needed in this case.

Alternatively, we could leave the horizon controls out of create, display the data the forecasting model generates along with the ground truth in the result screen, and then provide a "forecast" button that would then allow the user to set the horizon, and generate the forecast to be displayed in the following forecast exploration page (analogous to the flow from #1369).  

The latter seems like its forcing unnecessary steps into the process, but it is consistent to what would have to happen for classification/regression.

@slangevi @djonker - thoughts?

### LINKED ISSUE #1812: Timeseries does not need error range slider:
The error range slider shouldn't be displayed for timeseries data:

image

YohannParis:
closes #1812 
- added a test to **not** display the `ErrorThresholdSlider` on timeseries.

closes #1370 
- added a small test `isTimeseries` to check if a dataset is one on the routeGetter.
- updated the **PredictRequestMsg** interface to make some values optionals.
- updated the `ResultSummaries` and `SaveModel` components to display **Forecast** instead of **Apply Model** in the case of a timeseries.
- created the `ForecastHorizon` component to send a prediction request to the TA2.
- Added an error message.

Screen Shot 2020-07-28 at 14 24 33


phorne-uncharted:
My only concern would be with how the user specifies the interval length. If it is a datetime timeseries, that interval size needs to represent seconds or some other measurement. Regardless of what it represents, it is pretty ambiguous right now. Is it 5 days? 5 weeks? 5 hours?

YohannParis:
> My only concern would be with how the user specifies the interval length. If it is a datetime timeseries, that interval size needs to represent seconds or some other measurement. Regardless of what it represents, it is pretty ambiguous right now. Is it 5 days? 5 weeks? 5 hours?

Yes, that's why it is a draft so far. That's my next step to let the user decide the type of interval. Then to transform that value in seconds for datetime, or keep it an absolute integer.

YohannParis:
Based on the screenshot, please let me know if the phrases make sense and if they could be better written.

PR #1832, opened 2020-07-28
Remove call to geo referencing

### LINKED ISSUE #1815: Remove call to geo referencing
The geo-referencing pipeline isn't working properly on the cluster, and it is low priority for the time being.  We need to remove the call that initiates the geo-ref lookup on the client side so that we aren't introducing error state back into the application.

YohannParis:
closes #1815 

- Remove call to `datasetActions.geocodeVariable()` when a user change the type of a variable.
- Disable `datasetActions.geocodeVariable()` to return `null`.
- Remove the icon from the **StatusSidebar** component.

### DIFF
--- a/public/components/TypeChangeMenu.vue
+++ b/public/components/TypeChangeMenu.vue
@@ -340,12 +340,17 @@ export default Vue.extend({
           type: type
         })
         .then(() => {
+          /* TODO
+           * Disabled because the current solution is not responsive enough:
+           * https://github.com/uncharted-distil/distil/issues/1815
           if (isLocationType(type)) {
             return datasetActions.geocodeVariable(this.$store, {
               dataset: dataset,
               field: field
             });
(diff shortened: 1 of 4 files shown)


public/components/TypeChangeMenu.vue
```diff
@@ -226,7 +226,7 @@ export default Vue.extend({
         this.type === this.originalType && // we haven't changed the type (check from server)
         !this.isColTypeReviewed && // check if user ever reviewed the col type (client)
         this.hasSchemaType &&
-        this.schemaType.type != "unknown" && // don't flag for check when the schema type was unknown (which is the base type)
+        this.schemaType.type !== "unknown" && // don't flag for check when the schema type was unknown (which is the base type)
```
YohannParis, line 229: This was giving me a TS lint issue.

PR #1850, opened 2020-07-30
Support Expand/Collapse on target facet

### LINKED ISSUE #1841: Support Expand/Collapse on target facet
If a timeseries facet is selected as the target it doesn't show in expanded mode, which is correct.  The user should still have access to the type change menu to toggle the expand/collapse mode though.

YohannParis:
closes #1841 

I enable the type change on the TargetVariable component, it was on the sub-title, so I just moved it back to the facet.
I removed the title that was repeated in the sub-title and clean up a bit some HTML to make components easier to read.
Remove the **explode** commande in the TypeChange component on the select training features page.

Screen Shot 2020-07-30 at 16 40 58

### DIFF
--- a/public/components/TargetVariable.vue
+++ b/public/components/TargetVariable.vue
@@ -1,13 +1,13 @@
 <template>
-  <div v-bind:class="{ included: includedActive, excluded: !includedActive }">
+  <div :class="{ included: includedActive, excluded: !includedActive }">
     <variable-facets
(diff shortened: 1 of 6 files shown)


cdbethune:
Can you disable Explode in this case?  I'm not sure what that would do when the target is selected?

YohannParis:
Made the change, and update the main comment.

PR #1853, opened 2020-07-30
Display empty lines instead of 0 for missing values in timeseries

### LINKED ISSUE #1706: Render missing timeseres data differently than zeroed
We currently handle missing timeseries data by giving it a value of zero and rendering it as a such.  We should allow for the missing data to be rendered  in some distinguishing fashion.

YohannParis:
closes #1706 

Change backend to send `null` value instead of `0` if data is missing.
Added a **void** sparkline to illustrate the missing data. The strokes are so thin than using a gradient will be indistinguishable from real data.

### Sparkline
**Before** and **After**
Screen Shot 2020-07-30 at 16 07 56

### Line Chart
**Before**
Screen Shot 2020-07-30 at 16 20 19

 **After**
Screen Shot 2020-07-30 at 16 18 52

### DIFF
--- a/api/model/storage/postgres/timeseries.go
+++ b/api/model/storage/postgres/timeseries.go
@@ -244,7 +244,7 @@ func (s *Storage) FetchTimeseries(dataset string, storageName string, timeseries
 	where := fmt.Sprintf("WHERE %s", strings.Join(wheres, " AND "))
 
 	// Get count by category.
-	query := fmt.Sprintf("SELECT timeline.TimeStamps, COALESCE(filteredEvents.Counts, 0) FROM "+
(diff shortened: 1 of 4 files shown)


YohannParis:
I fill like we are not able to see as well values that are alone from a pack (the spike in the before images).

cdbethune:
Yeah, I think if want to go this way then we need to thicken the stroke up and / or use a higher contrast colour.  Maybe I can flip a screen shot to Pascale and see what she thinks.

cdbethune:
Pascale's suggestion:

image

YohannParis:
Yep that's I thought.
This one is going to be fun :)

YohannParis:
Updated with new scoring display:
Screen Shot 2020-07-31 at 14 44 41

Screen Shot 2020-07-31 at 14 44 33

PR #1869, opened 2020-08-05
Move the Test split threshold to a config env variables.

### LINKED ISSUE #1867: Train Test Split Ratio
The ratio between train and test datasets is currently using a constant defined in `api/compute/dataset_persist.go` but should instead be an environment variable to allow for variable splitting.

YohannParis:
closes #1867

### DIFF
--- a/api/routes/timeseries_forecast.go
+++ b/api/routes/timeseries_forecast.go
@@ -22,7 +22,6 @@ import (
 	"goji.io/v3/pat"
 
 	"github.com/uncharted-distil/distil/api/compute"
-	"github.com/uncharted-distil/distil/api/env"
 	api "github.com/uncharted-distil/distil/api/model"
 )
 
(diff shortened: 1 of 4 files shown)


api/routes/timeseries_forecast.go
```diff
@@ -114,7 +115,12 @@ func TimeseriesForecastHandler(metaCtor api.MetadataStorageCtor, dataCtor api.Da
 		}
 
 		// Recompute train/test split info for visualization purposes
-		split := compute.SplitTimeSeries(timeseries.Timeseries, compute.TimeSeriesTrainTestSplitThreshold)
+		config, err := env.LoadConfig()
```
phorne-uncharted, line 118: Instead of loading the config in the route, pass it in as a parameter (either the whole config struct or the split threshold). The cluster route handler has as parameter the whole config struct. Quite a few routes have specific config parameters.

YohannParis: Done

PR #1875, opened 2020-08-06
Result fetching limit by type

### LINKED ISSUE #1871: Result Fetching Limit By Type
Right now, the client will always pull back at most 100 rows of data from the server regardless of what type of data is being displayed and how it is being displayed.

The client should request varying amounts of rows based on the type of component being used for display. More rows of data need to be pulled when displaying remote sensing data to provide better coverage to the user. The exact amount of rows to pull back should be made easily changeable as the consequences of pulling more than 100 rows get clearer.

YohannParis:
closes #1871 

Created a new component that fetches new result data using the `size` filter.

The issue is this `size` is used to fetch both **included** and **excluded** results. So if a highlight exist, both included and excluded will return the number of results selected by the new component.

Screen Shot 2020-08-06 at 16 59 15
Screen Shot 2020-08-06 at 16 59 10


YohannParis:
I think I will change the design of the component to be a drop down slider, and display it where the number of results is.

To solve the **included** and **excluded** issue I will add boolean props to the component to fetch the right data.

YohannParis:
- Did a lot of clean-up.
- Change where the result size is displayed to account for **included** and **excluded** result-data-slots and change their size indivudually.

Screen Shot 2020-08-07 at 17 00 00

YohannParis:
I'm waiting to have some predictions capabilities to add this to the prediction component.

cdbethune:
The one thing to make sure of here is that the defaults are appropriately set.  We want to pull down more data when we're dealing with maps because we know that there aren't any costly secondary fetches.  If we're dealing with table data and we allow the row count to be turned up too high, the user could end up fetching a very large number of images or timeseries, which a) take a while to pull down and b) place a higher computational burden on the client.

YohannParis:
Add the same behaviour to the prediction page.

> The one thing to make sure of here is that the defaults are appropriately set. We want to pull down more data when we're dealing with maps because we know that there aren't any costly secondary fetches. If we're dealing with table data and we allow the row count to be turned up too high (or use a default that's too high) the user could end up fetching a very large number of images or timeseries, which a) take a while to pull down and b) place a higher computational burden on the client.

I completely agree! I can keep `100` for everything else but remote sensing that we set at `500`?
Do you think there is a way to know from a dataset a good amount of data?

cdbethune:
Can we enable this in the create model view as well?

YohannParis:
> Can we enable this in the create model view as well? 

Sure, and I will implement the improvements from @phorne-uncharted

YohannParis:
I added a [comment](https://github.com/uncharted-distil/distil/pull/1881#issuecomment-672214404) on @phorne-uncharted PR to be able to add this feature to the selection data slot component.

PR #1876, opened 2020-08-07
Feature Count And List

phorne-uncharted:
The client was counting every variable in the dataset and listing them all. However, there are quite a few variables that are not available to the user (grouping related ones, or clustering results) that should not count in the dataset summary.

This filters variables to only consider ones that have a data role. It isn't quite perfect since we still have issues with variables that get folded into groups (like band for remote sensing).

### DIFF
--- a/public/components/DatasetPreview.vue
+++ b/public/components/DatasetPreview.vue
@@ -161,6 +161,9 @@ export default Vue.extend({
     formatBytes(n: number): string {
       return formatBytes(n);
     },
+    filterVariablesByFeature(variables: Variable[]): Variable[] {
+      return filterVariablesByFeature(variables);
+    },
     setActiveDataset() {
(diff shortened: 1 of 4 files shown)


public/components/DatasetPreview.vue
```diff
@@ -13,7 +13,10 @@
         <i class="fa fa-table"></i> <b>Dateset Name:</b>
         {{ dataset.name }}
       </a>
-      <a class="nav-link"><b>Features:</b> {{ dataset.variables.length }}</a>
+      <a class="nav-link"
+        ><b>Features:</b>
+        {{ dataset.variables.filter(v => v.distilRole === "data").length }}</a
```
YohannParis, line 18: Usually, we create a **computed** function to not bloat the HTML with code. But I don't think it's an issue here.

PR #1881, opened 2020-08-11
Result filtered count

phorne-uncharted:
Added a number of filtered rows to the results so that the user can know how many rows in total there are with the selected filter.


YohannParis:
So this mean I will be able to provide a more accurate max number for the **included** or **excluded** results?

YohannParis:
Can we do this feature for `distil/data/`?

PR #1898, opened 2020-08-28
Adding Min/Max/Mean to Timerseries tables

### LINKED ISSUE #1678: Include min/max columns for timeseries table view
We should include min / max (mean?) columns for the timeseries table view so make it easier to see banding across series.

YohannParis:
closes #1678 

For the moment this is only added to the **SelectDataSlot** component, on the **Create** page.

If the goal is to make this available on **ALL** Timeseries table, then I will add the Min/Max/Mean calculation into the SQL request. Which will make things tidier.

### DIFF
--- a/api/model/storage/postgres/timeseries.go
+++ b/api/model/storage/postgres/timeseries.go
@@ -172,7 +172,7 @@ func (s *Storage) parseDateTimeTimeseriesForecast(rows pgx.Rows) ([]*api.Timeser
 }
 
 // Calculate the Min, Max, and Mean of a list of TimerseriesObservation
-func getMinMaxMean(timeseries []*api.TimeseriesObservation) (api.NullableFloat64, api.NullableFloat64, api.NullableFloat64) {
+func getMinMaxMean(timeseries []*api.TimeseriesObservation) (float64, float64, float64) {
 	min := math.Inf(1)
 	max := math.Inf(-1)
 	sum := float64(0)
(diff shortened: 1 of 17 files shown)


public/components/SelectDataTable.vue
```diff
@@ -128,7 +128,7 @@ export default Vue.extend({
   filters: {
     /* Display number with only two decimal. */
     cleanNumber(value) {
-      return parseFloat(value).toFixed(2);
+      return _.isNumber(value) ? value.toFixed(2) : "—";
```
YohannParis, line 131: Changes non available data to display a dash instead of **NaN**.
I choose a dash to make clear that we do not have any data, instead of missing data, as those are calculated information.

YohannParis: Screen Shot 2020-09-01 at 15 38 46

YohannParis:
I'm going to wait only I can test the changes I made for prediction page. But it should be good for review at least.

api/model/storage/postgres/timeseries.go
```diff
@@ -171,6 +171,39 @@ func (s *Storage) parseDateTimeTimeseriesForecast(rows pgx.Rows) ([]*api.Timeser
 	return points, nil
 }
 
+// Calculate the Min, Max, and Mean of a list of TimerseriesObservation
+func getMinMaxMean(timeseries []*api.TimeseriesObservation) (api.NullableFloat64, api.NullableFloat64, api.NullableFloat64) {
```
cdbethune, line 175: As a point of Go style - you would probably just want to return floats here, and then defer the NaN checks and use of the NullableFloat64 to the route handler.  The nullable type is only defined and used to support JSON serialization so you wouldn't want it permeating through other parts of the code.

cdbethune (approved):

PR #1903, opened 2020-09-01
Ingest create geometry

phorne-uncharted:
Made use of PostGIS features for remote sensing datasets. The coordinates are now stored as geometries, and queries use relevant functions to pull back data. The geobounds filter replaces the bivariate filter, and the remote sensing group hides the vector based coordinates field using instead the new geobounds based field.

NOTE: queries currently use WITHIN as operation to determine bucket contents for summaries. This means that tiles crossing bucket boundaries do not get counted. The INTERSECT operation would double count those same tiles. I am not sure which approach is best (I believe it currently does the equivalent of INTERSECT).

This PR also has minor fixes found while testing, including how the original variable name was being used.


api/dataset/satellite.go
```diff
@@ -82,20 +82,32 @@ type Point struct {
...
+		b.pointToString(b.LowerLeft, " "),
+		b.pointToString(b.UpperLeft, " "),
+		b.pointToString(b.UpperRight, " "),
+		b.pointToString(b.LowerRight, " "),
+		b.pointToString(b.LowerLeft, " "),
+	}
+	return fmt.Sprintf("POLYGON((%s))", strings.Join(coords, ","))
+}
+
+func (b *BoundingBox) pointToString(point *Point, separator string) string {
```
cdbethune, line 105: In case you hadn't run into this before - you could this a [Stringer](https://tour.golang.org/methods/17).

api/model/storage/postgres/request.go
```diff
@@ -67,8 +67,8 @@ func (s *Storage) PersistRequestFilters(requestID string, filters *api.FilterPar
 			if err != nil {
 				return errors.Wrap(err, "failed to persist numerical filter")
 			}
-		case model.BivariateFilter:
-			_, err := s.client.Exec(sql, requestID, filter.Key, model.BivariateFilter, filter.Mode, 0, 0, filter.Bounds.MinX, filter.Bounds.MaxX, filter.Bounds.MinY, filter.Bounds.MaxY, "", "")
+		case model.BivariateFilter, model.GeoBoundsFilter:
```
cdbethune, line 70: Can BivariateFilter be completely removed from the codebase or is it still used somewhere?

phorne-uncharted: I dont think it is used somewhere, so if that means it should be removed then so be it. Can you think of a use case where we could want that kind of filtering?

cdbethune: I guess if we wanted to do general 2D plots instead of geo?

PR #1906, opened 2020-09-02
Pre-featurized clustering

### LINKED ISSUE #1890: Pre-featurization of image data
In order to alleviate run time performance issues, we should be running the remote sensing featurizer over satellite data at the time of ingest, and use the pre-featurized data for processing tasks.  An initial implementation of this is available by setting the `FEATURIZATION_ENABLED` flag, but not heavily tested to date.

cdbethune:
fixes #1890

Pre-featurized data was working with the classification task, but clustering support needed to be added in.  The general clustering pipeline wasn't used because we want to cluster only the image related fields.


api/task/target_rank.go
```diff
@@ -16,63 +16,55 @@
...
-	if err != nil {
-		return nil, errors.Wrap(err, "unable to create ranking pipeline")
-	}
+	// // create & submit the solution request
+	// pip, err := description.CreateTargetRankingPipeline("roger", "", target, features)
+	// if err != nil {
+	// 	return nil, errors.Wrap(err, "unable to create ranking pipeline")
+	// }
+
+	return map[string]float64{}, nil
```
phorne-uncharted, line 32: Did you mean to leave this whole thing commented out?

cdbethune: No, I will re-enable it and re-test.

Issue #1755, opened 2020-07-14
Any summary action results in all summaries being refetched


cdbethune:
The _complete_ set of feature summaries is being fetched twice any time a facet-related action takes place.  For instance, in the model creation screen, adding or removing a feature will result in ALL feature summaries being fetched twice each time you trigger the action.  Clicking on a facet to highlight will likewise cause all feature summaries to re-fetch twice.  This shouldn't be necessary - once a summary has been fetched in a given view, it shouldn't need to be fetched again.  It should also be noted that facet pagination is no longer serving its intended purpose of managing summary fetches, as the full set of summaries are being fetched constantly.

cdbethune:
It looks like the summary fetches are cached, so for the most part we don't refetch data unnecessarily.  The one outstanding optimization is respecting the current page when we are filtering/highlighting.  When the user filters, it necessarily cause a refetch of facet data as the baseline vs. filter data is computed by the DB.  Paging was originally added to ensure that when filtering takes place, we only need to fetch data for those in the active page.  Otherwise, a filtering action on a dataset with 100 variables results in 100 fetches / updates.

karlolsonuc:
Looking at this, it seems we do nothing to respect the paging query strings anywhere beyond using the information there to reload on to the same page should we hard reload or direct navigate there. I think the approach will be to read the page info of various sorts from the query string to build a filter for the list of variables just before we make the included/excluded variable summary fetches which currently from any path taking that action do no filtering based on the page information in the route. We'll probably want a fall back of no filtering at all in the case of nothing defined rather than no fetches at all.

karlolsonuc:
Ah, I think we don't currently respect that because the variable facets component leans on the full list of summaries existing, so introducing pages into our current caching setup would essentially only pull a page at a time, but that then only displays that singular page of items. Even if you work around to expect the full list variables, the cache hits weird states as things are filtered, added and removed. Sometimes the filter update everything, sometimes they don't as there's a false cache hit.

PR #1907, opened 2020-09-03
#1755 - cache the summaries retrieved by action while only loading what the current page view needs.

karlolsonuc:
Previously, we leaned on every variable summary being loaded for change in state so that everything would stay updated by brute force, even though for any given view, we just need what's on screen (anything hidden by pagination shouldn't load until the page changes to reduce network calls and so we don't load data the user never sees because they changed filtering before changing the page.) This meant shifting from loading every variable in the dataset to only loading what's on screen given the pagination and active target variables, and saving that in a 2-deep dictionary keyed by variable name then current state as derived from the route and hashed to md5. This required changes to the store, getters, mutators and so on, as well modifications to the front end to expect those dictionaries instead filtering down the full array of information. This also meant changing our variable summary components pagination behavior as well, then changing any component or view that consumed that component to provide state in the expected way now.


public/util/data.ts
```diff
@@ -271,6 +273,22 @@ export function fetchResultExemplars(
   return new Promise((res) => res());
 }
 
+export function minimumRouteKey(): string {
```
YohannParis, line 276: It might be worth adding a comment explaining the purpose of this function.
This look like a major improvement, and the name `minimumRouteKey` is not really clear.

karlolsonuc: Will do.

PR #1918, opened 2020-09-10
added front end support for saving predictions

Zac-hills:
Added modal on prediction screen where user can download their predictions.
The modal contains a text input field where the user can define the name for the file (defaults to predictions.csv).
Added a fetchExportData function that calls the backend which generates the csv data.

### DIFF
--- a/public/components/PredictionSummaries.vue
+++ b/public/components/PredictionSummaries.vue
@@ -24,23 +24,22 @@
         </div>
       </div>
     </div>
-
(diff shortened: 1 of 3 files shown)


karlolsonuc (approved):
Looks good, though I'll note I didn't see the weighting information in the export. I don't know if we need that or not, so I'll leave final ok & merge to @cdbethune, but otherwise, the code looks fine and works fine on my local.

Issue #1880, opened 2020-08-10
FixedHeaderTable no longer needed


cdbethune:
`FixedHeaderTable` was needed because the Vue Bootstrap table didn't support a header that row that was always visible while the table contents scrolled.  Our own implementation of this functionality is fairly performance intensive, causing a noticeable lag when scrolling the table around with wide datasets or adding/removing facets.  The vue bootstrap component now supports a `sticky-header` property, which implements this feature internally, and seems to show no similar peformance hit.  We should drop `FixedHeaderTable` and use the native features of the table.

PR #1921, opened 2020-09-11
the fixed-table-header was hiding information if it went beyond 200px…

### LINKED ISSUE #1909: Fixed Header Table auto width caculation can bug
Because the component recalculates the header and cell width at each change of values, could be hundreds of times, the width can get messed up.

The following screenshot: the component calculates the mean's width to be 50px, but it gives the leftover space to the last column. When the Sparkline chart is rendered, the component keeps the mean column reference's excess size.

Screen Shot 2020-09-03 at 11 47 31

Zac-hills:
…. I opted to allow no limit and use a scrollbar. All information can now be seen, if the information is smaller than the entire table size the residual space is evenly distributed among the cells.
closes #1909

### DIFF
--- a/public/components/FixedHeaderTable.vue
+++ b/public/components/FixedHeaderTable.vue
@@ -35,17 +35,34 @@ export default Vue.extend({
       }
 
       const headTargetCells = [];
+      const bodyTargetCells = [];
       const minCellWidth = 200;
       const evenCellWidth = Math.ceil(
         this.tbody.clientWidth / theadCells.length
       );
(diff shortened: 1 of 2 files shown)


YohannParis:
This does solve the issue, but now the table does not fit into it's container.
I think that's why you added the `overflow-x: scroll`.

We want to keep all the table in one view if possible, and to keep the `max-width: 200px` for cell which do not have their own  `min-width` already set.

cdbethune:
Addressed by #1880

PR #1923, opened 2020-09-11
1899 badge highlight names

### LINKED ISSUE #1899: Highlight / Filter Badge Names Incorrect
Highlights / filter badge names are currently using the key (name) property of features. They should instead use the display name of features to match what gets used in the facets.

Zac-hills:
closes #1899 

Badge now displays the display name rather than the database key.


karlolsonuc (approved):
Looks good, I'll merge it back.

PR #1925, opened 2020-09-11
#1880 Fixed Header Removal

### LINKED ISSUE #1880: FixedHeaderTable no longer needed

karlolsonuc:
Closes #1880 - Leveraged new sticky-header prop for bootstrap-vue b-table to deprecate the slow fixed header table component/b-table wrapper, including removing the component from the code base and removing helper code in consuming components that nudged resizing actions in fixed table header instances. Added relevant css for tables to main.css as it the same styling for all tables with the class distil-table. Confirmed that it works fine with our more complex table cases including sparklines, images and weighting. This should be easy to restyle going forward as well when needed, and no longer has the O(nm) resizing overhead of the fixed-header-table.

Also:
- Caught all the route errors that previously just went straight to the console as errors and downgraded them to warnings. Longer term, we may want to look at why/when we're getting duplicate navigation.
- Fixed an issue with how the prediction export saveName was bound to the v-model that was causing an error.
- Fixed the z-index of the join table modal as it was hiding underneath the sidebar, making the far side of the modal unreachable.


public/styles/main.css
```diff
@@ -236,3 +236,29 @@ html {
   margin: 1em 0;
   padding-bottom: 0.25rem;
 }
+
```
YohannParis, line 239: Nice, lovely and clean.

public/components/SelectDataTable.vue
```diff
@@ -1,79 +1,77 @@
 <template>
-  <fixed-header-table ref="fixedHeaderTable">
-    <b-table
-      bordered
-      hover
-      small
-      :items="items"
-      :fields="tableFields"
-      @sort-changed="onSortChanged"
```
YohannParis, line 9: I guess we don't need to sort the table with the new <b-table> component?

public/components/ResultsDataTable.vue
```diff
@@ -1,123 +1,122 @@
...
+  <b-table
+    bordered
+    hover
+    small
+    :items="items"
+    :fields="tableFields"
+    :sort-by="errorCol"
+    :sort-compare="sortFunction"
+    @row-clicked="onRowClick"
+    @sort-changed="onSortChanged"
```
YohannParis, line 11: Here you kept the sorting

public/components/JoinDataTable.vue
```diff
@@ -1,51 +1,48 @@
...
-      :fields="emphasizedFields"
-      @sort-changed="onSortChanged"
-      @head-clicked="onColumnClicked"
+  <b-table
+    bordered
+    hover
+    small
+    :items="items"
+    :fields="emphasizedFields"
+    @sort-changed="onSortChanged"
```
YohannParis, line 8: Here you kept the function to sort, but the method is gone.

YohannParis (changes requested):

public/components/SelectDataTable.vue
karlolsonuc: Again, this was used to kick off fixed table state management.

public/components/ResultsDataTable.vue
karlolsonuc: I need to see if I should've removed it here too.

karlolsonuc: Ah, here because we pass in a special sort which in turn changes what sort it uses based on sort column, we need to listen to event change to update that sort state to pass that information back in the sort function.

public/components/JoinDataTable.vue
karlolsonuc: Ah , wait I see what you mean, I need to cut that listener.

YohannParis (approved):

PR #1926, opened 2020-09-14
added support for disabling view type radio buttons if there is no tr…

### LINKED ISSUE #1730: Removing the geocoordinates with remote sensing data does not remove the GeoPlot view.
If a remote sensing feature is selected the `GeoPlot.vue` is available for the user to view the data.
But it doesn't work if the geocoordinates data is not added.
The `GeoPlot.vue` should be available based on geocoordinates features, not remote-sensing ones.

bug

Zac-hills:
closes #1730 

Added support for disabling view-type-toggles if there is no training data for it to use.
Added a getter for routes that can check if geo_coordinate data is present. "uses training variable from route, should be changed to task variable once the backend supports it"
Remove some old prefix filtering in the filterName for FilterBadges(because we use the displayname now there is no need for filtering and making it look pretty)

### DIFF
--- a/public/components/FilterBadge.vue
+++ b/public/components/FilterBadge.vue
@@ -55,9 +55,6 @@ export default Vue.extend({
   computed: {
     filterName(): string {
       const type = getVarType(this.filter.key);
-      if (isClusterType(type)) {
-        return removeClusterPrefix(this.filter.key);
-      }
       return this.filter.displayName;
     },
     NUMERICAL_FILTER(): string {
--- a/public/components/FilterBadge.vue
+++ b/public/components/FilterBadge.vue
@@ -42,7 +42,7 @@ import {
   TEXT_FILTER,
 } from "../util/filters";
 import { clearHighlight } from "../util/highlights";
-import { getVarType, isClusterType, removeClusterPrefix } from "../util/types";
+import { removeClusterPrefix } from "../util/types";
 
 export default Vue.extend({
   name: "filter-badge",
@@ -54,7 +54,6 @@ export default Vue.extend({
 
   computed: {
     filterName(): string {
-      const type = getVarType(this.filter.key);
       return this.filter.displayName;
     },
     NUMERICAL_FILTER(): string {
--- a/public/components/GeoPlot.vue
+++ b/public/components/GeoPlot.vue
@@ -302,7 +302,7 @@ export default Vue.extend({
       const longitudes = [];
 
       const areas = this.dataItems.map((item) => {
-        const imageUrl = item.group_id.value;
+        const imageUrl = this.isRemoteSensing ? item.group_id.value : null;
         const fullCoordinates = item.coordinates.value.Elements;
         if (fullCoordinates.some((x) => x === undefined)) return;
 
@@ -354,6 +354,9 @@ export default Vue.extend({
 
     isRemoteSensing(): boolean {
       return routeGetters.isRemoteSensing(this.$store);
+    },
+    isGeoSpatial(): boolean {
+      return routeGetters.isGeoSpatial(this.$store);
     },
 
     /* Base layer for the map. */
@@ -752,9 +755,11 @@ export default Vue.extend({
         }).$mount();
 
         // Add interactivity to the layer.
-        layer
-          .bindTooltip(tooltip.$el as HTMLElement)
-          .on("click", () => this.showImageDrilldown(imageUrl, item));
+        layer.bindTooltip(tooltip.$el as HTMLElement).on("click", () => {
+          if (this.isRemoteSensing) {
+            this.showImageDrilldown(imageUrl, item);
+          }
+        });
 
         // Add the rectangle to the layer group.
         this.poiLayer.addLayer(layer);
@@ -817,7 +822,7 @@ export default Vue.extend({
       this.createMap();
       this.clear();
 
-      if (this.isRemoteSensing) {
+      if (this.isGeoSpatial) {
         // Display areas and update them on zoom to be sure they are selectable.
         this.displayAreas();
         this.map.on("zoomend", () => this.displayAreas());
--- a/public/components/GeoPlot.vue
+++ b/public/components/GeoPlot.vue
@@ -754,8 +754,9 @@ export default Vue.extend({
         }).$mount();
 
         // Add interactivity to the layer.
+        layer.bindTooltip(tooltip.$el as HTMLElement);
         if (this.isRemoteSensing) {
-          layer.bindTooltip(tooltip.$el as HTMLElement).on("click", () => {
+          layer.on("click", () => {
             this.showImageDrilldown(imageUrl, item);
           });
         }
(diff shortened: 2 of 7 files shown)


public/components/FilterBadge.vue
```diff
@@ -55,9 +55,6 @@ export default Vue.extend({
   computed: {
     filterName(): string {
       const type = getVarType(this.filter.key);
-      if (isClusterType(type)) {
```
YohannParis, line 58: You can remove the function import as well

public/components/GeoPlot.vue
```diff
@@ -752,9 +755,11 @@ export default Vue.extend({
         }).$mount();
 
         // Add interactivity to the layer.
-        layer
-          .bindTooltip(tooltip.$el as HTMLElement)
-          .on("click", () => this.showImageDrilldown(imageUrl, item));
+        layer.bindTooltip(tooltip.$el as HTMLElement).on("click", () => {
+          if (this.isRemoteSensing) {
```
YohannParis, line 759: You should put the `isRemoteSensing` test outside the `on("click...`

public/components/SelectDataSlot.vue
```diff
@@ -1,6 +1,11 @@
 <template>
   <div class="select-data-slot">
-    <view-type-toggle has-tabs v-model="viewTypeModel" :variables="variables">
+    <view-type-toggle
+      has-tabs
+      v-model="viewTypeModel"
+      :variables="variables"
```
YohannParis, line 6: Because those two props (variables, and trainingVariables) are in the store, why just not access them in the component itself?

public/components/ViewTypeToggle.vue
```diff
@@ -124,6 +135,23 @@ export default Vue.extend({
 
       return (hasLat && hasLon) || hasGeocoord;
     },
+    hasTrainingGeoVariables(): boolean {
```
YohannParis, line 138: instead of `Array.filter().length > 0` you can use `Array.some` that returns a boolean.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

public/store/route/getters.ts
```diff
@@ -546,6 +546,13 @@ export const getters = {
     return task.includes(TaskTypes.REMOTE_SENSING);
   },
 
+  isGeoSpatial(state: Route): boolean {
+    // get tasks in route
+    const task = state.query.training as string;
+    // return if geospatial reside in the route hardcoded for now I dont believe there is an enum for it
+    return !!task && task.includes("__geo_coordinates");
```
YohannParis, line 553: You should use a `TaskTypes` property instead of string, like you said in your comment.
Like on line 546

public/components/facets/GeocoordinateFacet.vue
```diff
@@ -85,6 +85,7 @@ import {
   NUMERICAL_SUMMARY,
   RowSelection,
   SummaryMode,
+  TaskTypes,
```
YohannParis, line 88: What do you need this for? If you need it, it should be in the `components/ViewTypeToggle.vue `

public/components/ViewTypeToggle.vue
```diff
@@ -58,6 +60,7 @@ import {
   LATITUDE_TYPE,
   GEOCOORDINATE_TYPE,
   REMOTE_SENSING_TYPE,
+  GEOBOUNDS_TYPE,
```
YohannParis, line 63: You forgot to include the updated `util.types.js` file in your PR.

YohannParis (changes requested):

public/components/ViewTypeToggle.vue
Zac-hills: This is keeping consistent with hasGeoVariables. If we change hasTrainingGeoVariables. I would recommend changing the above function to.

public/store/route/getters.ts
Zac-hills: I agree. However, this is checking the training variable not the task variable. The problem right now is geospatial is not supported as a task and I believe the server keeps removing it from the URL. Originally I went with that approach but until it is supported by the server we cannot use it.

public/components/FilterBadge.vue
Zac-hills: Will do.

public/components/GeoPlot.vue
Zac-hills: Will do.

public/components/SelectDataSlot.vue
Zac-hills: That could be a potential addition. However, variables was an existing prop and I opted for keeping the code similar. It could be refactored.

YohannParis: I think you are right, maybe this component is meant to be reused on other pages (results, etc.)
So make sure that this prop has a default value or that if missing it will not cause errors.

public/components/ViewTypeToggle.vue
YohannParis: Please do.

We do not have proper coding style in this project, only Prettier.js. So if you see code that can be rewrite in a manner that is clearer, cleaner, or faster, update it.

We don't need to be consistent, apart of naming convention.

public/store/route/getters.ts
YohannParis: Talk to @phorne-uncharted, or @cdbethune, about adding it on the backend.
If they think it's not necessary, etc. that might help you refine this PR.

public/components/ViewTypeToggle.vue
Zac-hills: Will do.

public/components/SelectDataSlot.vue
Zac-hills: Will do.

public/components/facets/GeocoordinateFacet.vue
Zac-hills: removed.

public/components/ViewTypeToggle.vue
Zac-hills: The GEOBOUNDS_TYPE already existed in the master.

public/components/GeoPlot.vue
```diff
@@ -755,11 +754,11 @@ export default Vue.extend({
         }).$mount();
 
         // Add interactivity to the layer.
-        layer.bindTooltip(tooltip.$el as HTMLElement).on("click", () => {
-          if (this.isRemoteSensing) {
+        if (this.isRemoteSensing) {
```
YohannParis, line 757: Sorry, but I think we still want to see the tooltip, just not have the `.on("click", ...` event listener.
More like:
```
layer.bindTooltip(tooltip.$el as HTMLElement);
if (this.isRemoteSensing) {
    layer.on("click", () => { this.showImageDrilldown(imageUrl, item) });
}
```

YohannParis (changes requested):

YohannParis (approved):

PR #1934, opened 2020-09-16
Display TA2 API version number

### LINKED ISSUE #1342: Display ta2 version string in app landing page
We currently display distil version/build info at the bottom of the home screen.  We should add the version info we retrieve from the TA2 via the `hello` message to that string.

YohannParis:
closes #1342 

- Expect to receive the TA2 API version number instead of a nil when starting the TA2 server.
- Add the TA2 and TA3 version number on the bottom right of the search view and as a title attribute on the app logo. ( this change is not responsive)

Screen Shot 2020-09-17 at 12 00 48

Screen Shot 2020-09-17 at 11 48 18


YohannParis:
I would like to spend more time displaying the version number in a nicer way, but I need to have distil-compute recompile with the changes I made there: https://github.com/uncharted-distil/distil-compute/pull/142

main.go
```diff
@@ -152,8 +153,9 @@ func main() {
 		return err == nil
 	}
 	servicesToWait["ta2"] = func() bool {
-		err := solutionClient.Hello()
-		return err == nil
+		versionNumber := solutionClient.Hello()
+		ta2Version = versionNumber
+		return versionNumber.(string)
```
phorne-uncharted, line 158: I think I understand what you are trying to do here. You still want to have the return statement be `return err == nil` since the function needs to return a bool indicating whether it succeeded or not.

phorne-uncharted:
This will need to wait until the relevant compute PR (https://github.com/uncharted-distil/distil-compute/pull/142) is merged and then the mod file will need to be updated.

main.go
YohannParis: I am trying to test if the version number is a String. But now that I'm writing this I guess in a typed language this is an irrelevant test...

I want to test if we got an answer from the handshake.

cdbethune (changes requested):
There are existing `version` and `timestamp` strings that capture info about the TA3 server that are returned in the same config message that you've updated.  Those would have been displayed on the home screen, which has since been removed.  They should be displayed at the same location as the ta2 version string.  We should also use a very low contrast font colour for the string, and keep the font size quite small - its info for us, not users.

go.mod
```diff
@@ -28,18 +27,16 @@ require (
 	github.com/russross/blackfriday v2.0.0+incompatible
 	github.com/shurcooL/sanitized_anchor_name v1.0.0 // indirect
 	github.com/stretchr/testify v1.5.1
-	github.com/uncharted-distil/distil-compute v0.0.0-20200909155738-39c3281e4a53
+	github.com/uncharted-distil/distil-compute v0.0.0-20200917145207-f4b488934399
 	github.com/uncharted-distil/gdal v0.0.0-20200504224203-25f2e6a0dc2a
 	github.com/unchartedsoftware/plog v0.0.0-20170413154239-34d2bbd3c0a9
-	github.com/unchartedsoftware/witch v0.0.0-20200617171400-4f405404126f // indirect
```
YohannParis, line 34: @cdbethune should I be worried?
This branch works fine when I run `make watch`.

cdbethune (approved):

PR #1935, opened 2020-09-18
1866 postgres type validating

### LINKED ISSUE #1866: Type Updates Too Limiting
If we import a dataset from CSV, and machine learned classification (Simon) returns unknown as suggested type, then there is no way for the user to switch the feature type to numerical (or at least no obvious way). I think the decimal type should be added to the basic suggested types to at least easily allow for that transition.

Zac-hills:
closes #1866 

Added postgres type validation support for the columns
Added a new Provenance 'postgres-valid'
Ingests properly and pulls the new valid types to the client
Client needs a refactor


api/postgres/postgres.go
```diff
@@ -725,3 +746,65 @@ func ValueForFieldType(typ string, field string) string {
...
+		case dataTypeInteger: 
+			return true 
+		case dataTypeDate:   
+			return true  
+		default:
+			return false 
+	}
+}
+// IsColumnType can be use to check columns potential types
+func (d *Database)IsColumnType(tableName string, variable *model.Variable, colType string) (bool, error) {
```
phorne-uncharted, line 771: It may be better to put this in the mode/storage/postgres package somewhere and expose it via the storage interface

restart_docker_&_app.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
phorne-uncharted, line 1: did you mean to commit this file?

run.sh
```diff
@@ -22,6 +22,7 @@ export SOLUTION_COMPUTE_PULL_TIMEOUT=60000
 export DATAMART_URL_NYU=https://auctus.vida-nyu.org
 export FEATURIZATION_ENABLED=false # re-enable when column parser is optimized
 export CLUSTERING_KMEANS=true
+export SUMMARY_ENABLED=false
```
phorne-uncharted, line 25: I would leave this off the commit

run_services.sh
```diff
@@ -19,5 +19,5 @@ fi
 if [[ -z "${DATAMART_URL_NYU}" ]]; then
   export DATAMART_URL_NYU="https://auctus.vida-nyu.org"
 fi
-
+export SUMMARY_ENABLED="false"
```
phorne-uncharted, line 22: I would leave this off the commit

api/task/ingest.go
```diff
@@ -635,3 +637,52 @@ func getUniqueDatasetName(meta *model.Metadata, storage api.MetadataStorage) (st
 
 	return getUniqueString(meta.Name, datasetNames), nil
 }
+
+func verifyData(pg *postgres.Database, meta *model.Metadata, storage api.MetadataStorage, tableName string) {
```
phorne-uncharted, line 641: If possible, it may be better to put this function in the postgres storage package. Then the caller would use the data storage interface to call the function. The postgres storage already has the metadata storage as struct property so no need to pass anything else around.

api/task/ingest.go
```diff
@@ -635,3 +637,52 @@ func getUniqueDatasetName(meta *model.Metadata, storage api.MetadataStorage) (st
...
+					suggestedType := model.SuggestedType{Probability:0, Type:d3mTypes[k], Provenance:provenance}
+					ds.Variables[i].SuggestedTypes = append(ds.Variables[i].SuggestedTypes, &suggestedType)
+				}
+			}
+		}
+	}
+	// save changes
+	storage.UpdateDataset(ds)
+	
+	return
```
phorne-uncharted, line 687: As general comment, this function needs to add error handling in some way. If we go forward with the assumption that suggested types has the complete list, then this needs to be considered an integral part of the ingest process so any failure here needs to fail the ingest.

The function should probably have error as return type, and whenever something errors it should forward that error.

phorne-uncharted:
I think theres some code formatting inconsistencies going on. Whatever plugin used for go should format it properly, but if it doesn't you can always use `gofmt -s -w .` from the root directory to format all go files.

PR #1946, opened 2020-09-23
Fixed errors color with new facet function

### LINKED ISSUE #1931: Error facet should be red
Error facets are rendered in the default blue, instead of the red previously used.

image

Zac-hills:
closes #1931 

Added a function to facet.ts.
Responsibility is to change the dynamic variables surrounding the facet library.
Altered ResultGroup.vue to use the aforementioned function.


karlolsonuc (approved):
Great work.

PR #1949, opened 2020-09-24
#1901 performance fixes

### LINKED ISSUE #1901: Large table view renders very slowly
Using ACLED and the setting the table to display 2000 items when all features are being modelled causes the system to lock up for  10 - 20 seconds evertime the table content is changed (for instance, by selecting a category through a facet).  Profiling shows all of this time being spent in the bootstrap-vue table control `render` function.

karlolsonuc:
Closes #1901
Slow table render was a three fold issue:
1. Too many rows on screen as we were now able to show more rows than the top 100. Solved by bootstrap-vue pagination as skeleton/virtualized scrolling didn't seem to readily integrate with the bootstrap-vue table component. Still needed despite other improvements as rendering very large tables hits browser drawing slow downs.
2. Each facet previously needed to rescan all table rows to remap the d3m index representation of row selection in the URL query string back to the row data. Solved by adding and maintaining the current row selection in the VueX store, and referencing that directly from the store by the utility function that facets use for subSelection, resulting in the search to update the rowSelection only occurring once as the rowSelection changes, not for all visible facets after any change.
3. Too many listeners being set on the table data as it was built and used by the b-table component by virtue of API returns not being frozen or sealed either when being saved to the VueX store or when being returned from a mapping function that tapped that store, making any action involving that data take much longer, as VueX had to deep check to see if any access had caused a change to the underlying structure or it's values, even though almost all changes are posted back to the server and result in reloading the data. This was particularly bad with larger data sets because it incurred a number of very slow garbage collection steps.


public/components/JoinDatasetsForm.vue
```diff
@@ -198,7 +198,7 @@ export default Vue.extend({
         .then((tableData) => {
           this.pending = false;
           this.showJoinSuccess = true;
-          this.previewTableData = tableData;
+          this.previewTableData = Object.seal(tableData);
```
YohannParis, line 201: Sorry to be pedantic about this, but can you add a small comment on why you use `Object.seal()`. Usually, people use this method to make the object non-mutable, but here it's to circumvent the bootstrap-component.

public/store/dataset/mutations.ts
```diff
@@ -334,12 +335,21 @@ export const mutations = {
 
   // sets the current selected data into the store
   setIncludedTableData(state: DatasetState, tableData: TableData) {
-    state.includedSet.tableData = tableData;
+    state.includedSet.tableData = Object.freeze(tableData);
```
YohannParis, line 338: Same comment for `Object.seal()`, a small comment might avoid the future issue by someone trying to be smart and removing it.

public/components/JoinDatasetsForm.vue
karlolsonuc: I can add that really quick. In a sense here, it's also an issue of immutability, just for different reasons, but yes, I need to make it clear so it never is callously reverted.

karlolsonuc: Comments added 👍

YohannParis (approved):

PR #1950, opened 2020-09-24
Parquet result handling

phorne-uncharted:
Switched to uncompressed parquet writing due to TA2 library limitations.

Produce request has parquet uri as first exposed value type if working with a featurized dataset.


api/compute/solution_request.go
```diff
@@ -681,7 +683,11 @@ func (s *SolutionRequest) dispatchSolution(statusChan chan SolutionStatus, clien
 				break
 			}
 		}
-		produceSolutionRequest := createProduceSolutionRequest(produceDatasetURI, fittedSolutionID, outputKeys)
+		exposeType := []string{}
+		if s.useParquet {
+			exposeType = append(exposeType, "PARQUET_URI")
```
cdbethune, line 688: Should add a constant to `ta3ta2.go` for this string and use that instead.  The other value types all have defs there.

PR #1952, opened 2020-09-25
1751 ranking clean up

### LINKED ISSUE #1751: Ranking cleanup
The ranking / feature importance code is not very well organized or consistent by this point.  The terms "ranking" and "importance" are used interchangeably, and we store ranks in the variables themselves for non-results, when we should keep things in a separate map in the store.

Zac-hills:
Clean up surrounding the ranking code.
There is three rankings PCA, MI, and FI.
Moved the rankings from the variable and into a map.
Changed a lot of the helper functions in data.ts to reflect the above.
Fixed a bug where we would fetch FI too early and we would never display FI. 
Added Ranking fix in result.
closes #1751


karlolsonuc (approved):
Looks good, works on my local, merging back.

PR #1956, opened 2020-09-30
#1879 - Model Explanation Badge

### LINKED ISSUE #1879: Add a badge to show when a model has explanations available
We should add a badge next to the score for a model that indicates whether or not it has explanations available.  Currently we just don't show SHAP colouring, and swap the MI ranking values in for the missing feature importances.

karlolsonuc:
Closes #1879  - Added boolean weighted field to the return for predicted summaries with helper SQL query that checks if we return even one weighted row back, then read that field off to set hasExplanations computed in the result group component, with updates to the variable summary inferfaces in go and typescript to support the weighted field.

### DIFF
--- a/public/components/ResultGroup.vue
+++ b/public/components/ResultGroup.vue
@@ -332,7 +332,7 @@ export default Vue.extend({
 
     hasExplanations(): boolean {
       // waiting for explanation enum to be added to results summaries
-      return false;
+      return this.predictedSummary?.weighted;
     },
   },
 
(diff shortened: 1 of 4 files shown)


public/components/ResultGroup.vue
```diff
@@ -332,7 +332,7 @@ export default Vue.extend({
 
     hasExplanations(): boolean {
       // waiting for explanation enum to be added to results summaries
-      return false;
+      return this.predictedSummary?.weighted;
```
YohannParis, line 335: With [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) you can get a `undefined`, adding `!!` in front will make sure you get a boolean.

cdbethune (approved):

public/components/ResultGroup.vue
karlolsonuc: Good point will update.

PR #1965, opened 2020-10-06
Added route watcher to selectTraining view for cluster persistence. C…

### LINKED ISSUE #1958: Clustering / ranking state not correct
The clustering/ranking state is often incorrect.  When applied, the red dot on in the status panel should be cleared and the ranking / clustering should be applied to the data.   We've stored this in the route to make it sticky, but it doesn't seem to properly maintain state on refreshes or browser forward / back actions.

Zac-hills:
-Removed some unused imports from App.vue.
-Added route watch in SelectTraining view. (Used to update variables if clustering was added through browser navigation)
-Updated FetchVariableSummary to prioritize URL params over given function params if available.
-Added check to SideBar to only present red dot if data has not been clustered.
closes #1958

### DIFF
--- a/public/App.vue
+++ b/public/App.vue
@@ -12,10 +12,7 @@ import VueObserveVisibility from "vue-observe-visibility";
 import BootstrapVue from "bootstrap-vue";
 import NavBar from "./components/NavBar.vue";
 import store from "./store/store";
-import {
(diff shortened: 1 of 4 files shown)


karlolsonuc (approved):
Looks good to me.

PR #1966, opened 2020-10-06
1846 mi ranking result view

### LINKED ISSUE #1846: Verify MI ranking is running on the model result screen only when expected
MI ranking could end up kicking off from the result screen if we are displaying a model that doesn't provide feature importances.  If the ranking run was already done, the values should be available immediately.  If it wasn't done, then it is valid for the run to start.  Whether or not this is actually the case needs to be verified.

Zac-hills:
-Added helper function to data.ts with the purpose of checking if MI still needs to be executed.
-Added check to result view to check if MI needs to be run (this is an edge case where we load asaved model in that happens to not have MI stored).
closes #1846

### DIFF
--- a/public/util/data.ts
+++ b/public/util/data.ts
@@ -4,6 +4,7 @@ import _, { result } from "lodash";
 import Vue from "vue";
 import {
   D3M_INDEX_FIELD,
+  DatasetPendingRequestType,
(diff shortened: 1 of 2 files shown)


karlolsonuc (approved):
Looks good.

PR #1974, opened 2020-10-07
#1943 - Filter Pattern Facets From Display

### LINKED ISSUE #1943: Pattern facet shouldn't show up in remote sensing model create
After generating patterns, the pattern facet is showing up in the remote sensing facet set:

image

Since patterns are applied to the image facet, it shouldn't ever be displayed.

karlolsonuc:
Closes #1943 - Added a step in getVariableSummariesByState to filter variables with a "_cluster" prefix from display.

### DIFF
--- a/public/util/data.ts
+++ b/public/util/data.ts
@@ -580,15 +580,14 @@ export function getVariableSummariesByState(
   let currentSummaries = [];
 
   if (Object.keys(summaryDictionary).length > 0 && variables.length > 0) {
-    let sortedVariables = variables;
+    // remove any pattern cluster variables
+    let sortedVariables = variables.filter((sv) => {
+      return sv.colName.indexOf(CLUSTER_PREFIX) < 0;
+    });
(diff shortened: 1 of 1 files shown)


public/util/data.ts
```diff
@@ -584,7 +585,10 @@ export function getVariableSummariesByState(
       // prioritize FI over MI
       sortedVariables = sortVariablesByImportance(sortedVariables);
     }
-
+    // remove any pattern cluster variables
+    sortedVariables = sortedVariables.filter((sv) => {
```
YohannParis, line 589: I think that filtering should be done on line 583 when `sortedVariables` is created.
Like this we make sure that the cluster variables are not ranked by mistakes, etc.

YohannParis (approved):

public/util/data.ts
karlolsonuc: Good point.

PR #1981, opened 2020-10-14
Confidence Summary Route

phorne-uncharted:
Creation of the confidence summary route (`/distil/confidence-summary/:dataset/:results-uuid/:mode`) that has the same input and output structure as the correctness summary route. It creates 51 buckets in intervals of 0.02 with extrema [0,1].


karlolsonuc (approved):
Looks excellent!

PR #1983, opened 2020-10-15
Webgl geoplot

### LINKED ISSUE #1977: Add Geo interface capable of handling 50k tiles.

Zac-hills:
closes #1977 

Objective: Webgl Geoplot, capable of rendering large datasets.

Webgl Geoplot currently should support:
- OnHover display tile information in Toast in top right hand corner of map (upon moving toast should fade)
- OnClick trigger DrillDown (Should be the exact same as the previous Geoplot)
- FitToBounds on receiving new data the map should fit itself to the set making sure every tile is visible on screen (it should work with the number of displayed data drop down and GeoFacet highlighting)
- Quads should represent the color within the this.areas[].color
- Map should pan, zoom and load tiles

How it works:
- Single VBO for all the quads (this ensures that there is only one draw call at any scale of data)
- Two FrameBuffer Objects (one that contains the normal visual data that the user sees. Another that is never visually seen, it contains the ids of the quads for what is on screen)(https://webglfundamentals.org/webgl/lessons/webgl-picking.html)
- There is two render calls, one which renders normally and another which renders to the the id buffer (only because webgl 1 does not support multiple render targets in the fragment shader)


Zac-hills:
Added Yohann and Karl per our conversation during the meeting.

karlolsonuc (approved):
Looks good for this phase of the work.

YohannParis (approved):
Looks to work fine on my machine, and I need to learn more about WebGL and Lumo. But it looks clean enough for a first PR.

I spotted only one big issue, the selection tool does not work.
I know you are going to implement this in another PR, but for now you should hide the button.
Screen Shot 2020-10-15 at 16 38 42

karlolsonuc (approved):

PR #1985, opened 2020-10-16
Summarized geo tiles

Zac-hills:
- Added zoom states to geo plot. 
Peek 2020-10-16 13-52
- First state, is the clustered state where the geoplot renders the clustered tiles.
- Second state, is the normal tile rendering state where geoplot renders all of the tiles.
- First state has a click event that zooms into the center of the cluster to the zoom threshold. By reaching the zoom threshold it switches states to the normal tile rendering state.
- Added some support for no data being received (just render an empty map and when data is receieved figure out what state it should be in)
- Added a fix for the GeoCoordinateFacet, it was sending data as lnglat and needs to be sent as latlng.

little unhappy with the zoom function for lumo it seems linear and with a constant time to zoom.
Peek 2020-10-16 13-53
closes #1985


public/components/facets/GeocoordinateFacet.vue
```diff
@@ -320,11 +279,11 @@ export default Vue.extend({
             const feature = polygon(
               [
                 [
-                  [xCoord, yCoord],
-                  [xCoord, yCoord + ySize],
-                  [xCoord + xSize, yCoord + ySize],
-                  [xCoord + xSize, yCoord],
-                  [xCoord, yCoord],
+                  [yCoord, xCoord],
```
YohannParis, line 282: Can you explain in a small comment why you went from [x, y] to [y, x], which is unusual.

public/util/rendering/BatchQuadOverlayRenderer.js
```diff
@@ -214,7 +212,6 @@ export class BatchQuadOverlayRenderer extends WebGLOverlayRenderer {
     const cell = plot.cell;
     const proj = this.getOrthoMatrix();
     const scale = Math.pow(2, plot.zoom - cell.zoom);
-    const opacity = this.overlay.opacity;
```
YohannParis, line 217: Why removing the opacity? I think it makes it harder to know where the buckets are, specially in a use case where a whole region could be covered.

public/components/GeoPlot.vue
```diff
@@ -427,8 +490,62 @@ export default Vue.extend({
...
+            bucket.coordinates[0]
+          );
+          const point2 = this.renderer.latlngToNormalized(
+            bucket.coordinates[1]
+          );
+          const center = {
+            x: (point1.x + point2.x) / 2,
+            y: (point1.y + point2.y) / 2,
+          };
+          this.map.zoomToPosition(this.zoomThreshold, center); // zoom to the center of the cluster clicked. Zoom to the point where the state switches
```
YohannParis, line 541: Aren't you worried that on a smaller screen (i.e. laptop) that `this.zoomThreshold` might be too high and that the canvas might only show a partial bucket?

public/components/GeoPlot.vue
```diff
@@ -456,18 +573,31 @@ export default Vue.extend({
...
       // convert this.areas to quads in normalized space and add to overlay layer
-      const normalizedQuads = this.areaToQuads();
+      if (!this.bucketFeatures.length) {
+        this.currentState = this.tileState;
+      } else {
+        this.currentState = this.clusterState;
+        this.map.on(lumo.ZOOM_END, this.onZoom);
+      }
+      if (!this.bucketFeatures.length && !this.areas.length) {
+        this.map.add(this.overlay);
```
YohannParis, line 583: Why adding an overlay if there is no data?

public/components/GeoPlot.vue
```diff
@@ -511,11 +642,50 @@ export default Vue.extend({
...
     },
+    bucketsToQuads(): Quad[] {
+      const maxVal = this.maxBucketCount;
+      const minVal = this.minBucketCount;
+      const d = (maxVal - minVal) / BLUE_PALETTE.length;
+      const domain = BLUE_PALETTE.map((val, index) => minVal + d * (index + 1));
+      const scaleColors = scaleThreshold()
+        .range(BLUE_PALETTE as any)
+        .domain(domain);
+      const singleBuffer = [];
```
YohannParis, line 653: I remember you explaining that WebGL uses a buffer for rendering. Do you mind adding a comment explaining this? This constant is created purely for WebGL, and it's worth mentioning.

public/components/GeoPlot.vue
```diff
@@ -976,11 +1178,11 @@ export default Vue.extend({
...
-      const normalizedQuads = this.areaToQuads();
+      const quads = this.currentState.quads();
       // get bounds of quad set
-      const mapBounds = this.getBounds(normalizedQuads);
+      const mapBounds = this.getBounds(quads);
       // add the batched quads to a single layer on the overlay
-      this.overlay.addQuad(this.polygonLayerId, this.areaToQuads());
+      this.overlay.addQuad(this.polygonLayerId, quads);
       // fit map to the quad set
       this.map.fitToBounds(mapBounds);
```
YohannParis, line 1187: Why do you use `map.fitToBounds()` on `onNewData()` but not on `updateMap()`?

YohannParis (changes requested):
I tried with `bigearth_tiny` and `bigearth-tiny-sample-first-only` datasets and the buckets are way too big compare to the individual image quads.

new

public/components/GeoPlot.vue
```diff
@@ -148,7 +151,10 @@ interface Area {
   imageUrl: string;
   item: TableRow;
 }
-
+interface Bucket {
```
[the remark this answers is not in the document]
Zac-hills: Will do!

public/util/rendering/BatchQuadOverlayRenderer.js
Zac-hills: This is a const opacity that was being applied to every quad. The quads still support opacity in their VBO. Essentially what was happening in the shader was vbo.alpha * overlay.opacity. <- so the overlay.opacity is kind of redundant since we can control the opacity through the vertex data.

public/components/facets/GeocoordinateFacet.vue
Zac-hills: In leaflet it uses the latlng which is y,x. 
Screenshot from 2020-10-16 14-59-15

public/components/GeoPlot.vue
Zac-hills: The overlay still needs to be added. I should just move that above the if statement.

public/components/GeoPlot.vue
Zac-hills: onNewData gets triggered when new data is being passed in such as highlighting. The map fits the new data to screen to all is visible. updateMap is used when states change from cluster to tile and the map needs to update the renderer. I will add comments for the functions.

YohannParis (approved):

Issue #1421, opened 2020-01-20
Convert all numerical vars to timeseries when one is selected


cdbethune:
When we convert a given variable to a timeseries, any other numeric variables should be converted along with it using the time column and series ID keys.

PR #2005, opened 2020-10-21
1610 adding categorical feature update timeseries grouping

YohannParis:
This PR is my smallest iteration to get the job done. The groupings might need a cleaning at some point, as what I've done looks more like a hack than an enhancement.

- Create a `updateGrouping()` action that deletes the current grouping and create a new one.

- When the user is selecting features, if the target variable is a time series: all categorical variables **Add** button change to **Add to Timeserie** and will call `updateGrouping()` with current grouping information taken from the store.

- Clean up the **VariableGrouping** page.

### DIFF
--- a/public/components/AvailableTrainingVariables.vue
+++ b/public/components/AvailableTrainingVariables.vue
@@ -139,7 +139,7 @@ export default Vue.extend({
         const isCategorical: boolean = group.type === "categorical";
         if (this.isTimeseries && isCategorical) {
           // Change the meaning of the button as this action is different than the default one.
-          trainingElem.textContent = "Add to Timeserie";
+          trainingElem.textContent = "Add to Timeseries";
         }
 
         trainingElem.addEventListener("click", async () => {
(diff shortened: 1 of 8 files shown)


public/components/AvailableTrainingVariables.vue
```diff
@@ -94,30 +96,52 @@ export default Vue.extend({
...
         const trainingElem = document.createElement("button");
         trainingElem.className += "btn btn-sm btn-outline-secondary mb-2";
-        trainingElem.innerHTML = "Add";
+        trainingElem.textContent = "Add";
+
+        // In the case of a categorical variable with a timeserie selected.
+        const isCategorical: boolean = group.type === "categorical";
+        if (this.isTimeseries && isCategorical) {
+          // Change the meaning of the button as this action is different than the default one.
+          trainingElem.textContent = "Add to Timeserie";
```
cdbethune, line 142: Should be "Add to Timeseries"

cdbethune (changes requested):
In testing this, I found one other thing that may be useful - being able to remove elements in the same way they are added.  Note that there are overall stability issues, but they are being cleaned up as part of #1421.

Issue #1979, opened 2020-10-10
Allow user to select metrics in a simplified fashion


cdbethune:
We currently don't allow users to select metrics at all.  A simple approach would be to provide a dropdown in the model settings panel to allow them to select metrics from a list that are relevant given the problem type and target selection.  Better would be to capture how the user wants to evaluate the model, and determine the appropriate metric(s) to apply.  The values returned to user should then somehow be linked back to their intended evaluation criteria, rather than just providing them with a metric name that is likely meaningless to them.

PR #2011, opened 2020-10-23
#1979 select modelling metrics

karlolsonuc:
Added new distil/model-metrics route to api and consumed it in the front end to populate dropdown & descriptions in model settings that provides different metrics based on ML task type (broadly classification vs. regression, with special handlers for binary vs. multi-class classification.)

### DIFF
--- a/api/routes/model_metrics.go
+++ b/api/routes/model_metrics.go
@@ -1,0 +1,69 @@
+//
+//   Copyright © 2020 Uncharted Software Inc.
+//
+//   Licensed under the Apache License, Version 2.0 (the "License");
+//   you may not use this file except in compliance with the License.
+//   You may obtain a copy of the License at
+//
+//       http://www.apache.org/licenses/LICENSE-2.0
+//
+//   Unless required by applicable law or agreed to in writing, software
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+
+package routes
+
+import (
+	"net/http"
+	"strings"
+
+	"github.com/pkg/errors"
+	api "github.com/uncharted-distil/distil/api/model"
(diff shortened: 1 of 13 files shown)


api/routes/model_metrics.go
```diff
@@ -0,0 +1,75 @@
...
+			} else if strings.Contains(task, util.BinaryTask) {
+				taskMetrics = util.TaskMetricMap[util.BinaryTask]
+			} else {
+				taskMetrics = util.TaskMetricMap[util.ClassificationTask]
+			}
+		} else if strings.Contains(task, util.RegressionTask) || strings.Contains(task, util.ForecastingTask) {
+			taskMetrics = util.TaskMetricMap[util.RegressionTask]
+		} else {
+			taskMetrics = util.AllModelMetrics
+		}
```
phorne-uncharted, line 59: I want to make sure I get this if properly. For it to be a multi classification problem, the task keywords must contain `classification` AND `multiClass`? So if only `multiClass` is sent, this would use `util.AllModelMetrics`?

phorne-uncharted (approved):

api/routes/model_metrics.go
karlolsonuc: Yes, though we should never get into that state as we're always sending the classification task tag back as well.

api/util/metrics.go
```diff
@@ -0,0 +1,188 @@
...
+//
+//   Unless required by applicable law or agreed to in writing, software
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+
+package util
+
+const (
```
cdbethune, line 18: There are constant definitions in `ta3ta2.go` from `distil-compute` that overlap with these - definitely those for task kewyords and metric names.  We should re-use the existing definitions.

karlolsonuc: Okay replaced the duplicate consts and such with imports from compute where applicable.

PR #2013, opened 2020-10-26
Google maps base tiles

### LINKED ISSUE #1988: Google maps as base layer
We should have an option to run with google maps as our base layer, as we need access to satellite imagery for remote sensing work.  We should still allow for our existing base layer to be used as a config option.

Zac-hills:
- "TILE_REQUEST_URL": "https://server.arcgisonline.com/ArcGIS/rest/services/World_Imagery/MapServer/tile/{z}/{y}/{x}.png"
- Use the above environment variable to change the map of the GeoPlot. (The above url should work for the aerial images and it is free)
closes #1988


phorne-uncharted:
I think go formatting is slightly off. `gofmt -s -w .` should fix all files if nothing else works. There are also a couple of commented out lines that should probably be removed.

PR #2040, opened 2020-11-03
#2029 error color fix in facets

### LINKED ISSUE #2029: Histogram red doesn't match other uses
The red used in the histograms doesn't match the red used else for error representation:

image

karlolsonuc:
Fixes #2029. Just reused the colors from the table error styling - one line change.


Zac-hills (commented):
One thing that might be useful is moving all these colors into the color.ts file. I noticed a lot of hardcoded color values throughout the components.

Zac-hills (approved):

PR #2054, opened 2020-11-04
Use different folder for server ingest

### LINKED ISSUE #2042: Re-use of augmented directory for server ingest confusing
We've added support for importing datasets from a directory on the server side, currently it defaults to `augmented`.  This leads to some odd behavior, becuase when I ingest from my local machine, the dataset is added to `augemented`.  If I bring the import dialog up again, it now shows the dataset I just imported as a dataset availabe on the server to import.

YohannParis:
closes #2042

Created a new _public_ folder under `$D3MOUTPUTDIR`.

- Updated distil-compute to add a folder as a dataset source.
- New ENV variable `PUBLIC_SUBFOLDER` to specify the folder.
- Updated the _AddDataset_ component and the _importAvailableDataset()_ to use the folder.

### DIFF
--- a/api/env/config.go
+++ b/api/env/config.go
@@ -73,6 +73,7 @@ type Config struct {
 	PostgresPort                       int     `env:"PG_PORT" envDefault:"5432"`
 	PostgresRandomSeed                 float64 `env:"PG_RANDOM_SEED" envDefault:"0.2"`
 	PostgresUser                       string  `env:"PG_USER" envDefault:"distil"`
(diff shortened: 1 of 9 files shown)


phorne-uncharted (approved):

YohannParis:
One nice improvement to the _available datasets_ list would be to show an icon or status next to a dataset that was previously imported. For example:

Screen Shot 2020-11-04 at 14 30 46

PR #2055, opened 2020-11-04
#2035 facet colour flicker fix

### LINKED ISSUE #2035: Colour cycling / redraw on category click
When clicking on a category for the first time there is some type of colour cycling that occurs that is pretty distracting:

Peek 2020-11-02 15-41

It seems to stop the 2nd time a category is clicked.

karlolsonuc:
Closes #2035. Cut the facet colouring from main.css and set it dynamically in VariableFacets & ResultGroup based on current selection, filter and exclude state such that that colour cycle flickering should be avoided now as those colours shouldn't be defined when there's no relevant data for them. Also updated facets utility function applyColor to skip null entries so it can fall back to facet defaults as needed. Also, lots of tslint & prettier formatting clean up.


Zac-hills (commented):
It looks good. I see two things that might be helpful. The first, adding | null to the FacetColor properties since they can be null now. The second, possibly moving the color values to color.ts, just in case we decide to change the pallette down the road. Something like FacetColorPalette{Error: "#888888", ErrorHover:"#999999"}

karlolsonuc:
I like those ideas, I'll add them in.

Zac-hills (approved):

PR #2065, opened 2020-11-10
Color brightness correction

Zac-hills:
Peek 2020-11-10 15-35
- Added gainL control for the user to inspect images
- Added optional params to the Multiband_image route

closes #2065

### DIFF
--- a/api/routes/multiband_image.go
+++ b/api/routes/multiband_image.go
@@ -46,16 +46,11 @@ func MultiBandImageHandler(ctor api.MetadataStorageCtor) func(http.ResponseWrite
 		bandCombo := pat.Param(r, "band-combination")
 		paramOption := getOptions(r.URL.Path)
 		isThumbnail, err := strconv.ParseBool(pat.Param(r, "is-thumbnail"))
+		if err != nil {
+			handleError(w, err)
+			return
+		}
 		imageScale := util.ImageScale{}
-		if err != nil {
-			handleError(w, err)
-			return
-		}
-
-		if err != nil {
-			handleError(w, err)
-			return
-		}
 		// get metadata client
 		storage, err := ctor()
 		if err != nil {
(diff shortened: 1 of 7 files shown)


public/components/ImageDrilldown.vue
```diff
@@ -6,18 +6,38 @@
...
+        :item="item"
+      />
+      <div
+        class="image-container"
+        ref="imageContainer"
+        :style="{ '--IMAGE_MAX_SIZE': IMAGE_MAX_SIZE + 'px' }"
+      ></div>
+      <div class="slider-container">
+        <label class="slider-label">0.0</label>
+        <input
```
karlolsonuc, line 24: Maybe consider using the bootstrap-vue b-form input slider for (as seen in DataSize.vue) for style consistency but otherwise fine.

karlolsonuc (approved):

public/components/ImageDrilldown.vue
Zac-hills: I will add the change to use bootstrap slider.

api/routes/multiband_image.go
```diff
@@ -24,21 +25,33 @@ import (
...
 // MultiBandImageHandler fetches individual band images and combines them into a single RGB image using the supplied mapping.
 func MultiBandImageHandler(ctor api.MetadataStorageCtor) func(http.ResponseWriter, *http.Request) {
 	return func(w http.ResponseWriter, r *http.Request) {
 		dataset := pat.Param(r, "dataset")
 		imageID := pat.Param(r, "image-id")
 		bandCombo := pat.Param(r, "band-combination")
+		paramOption := getOptions(r.URL.Path)
 		isThumbnail, err := strconv.ParseBool(pat.Param(r, "is-thumbnail"))
 		imageScale := util.ImageScale{}
+		if err != nil {
```
phorne-uncharted, line 50: This should be put after the line that could error. Either the strconv call should be moved down or this moved up one.

PR #2068, opened 2020-11-11
added map tile set toggle

### LINKED ISSUE #2064: Add satellite/streetmap toggle to map view

Zac-hills:
closes #2064 
- Added map toggle for geoplot
map-toggle


public/components/GeoPlot.vue
```diff
@@ -58,6 +58,15 @@
         C
       </a>
     </div>
+    <div
+      class="map-toggle"
+      :class="{ active: isSatelliteView }"
+      @click="mapToggle"
+    >
+      <a class="cluster-icon" title="Change Map" aria-label="Change Map">
```
YohannParis, line 66: I would have made this a _button_ to be semantically correct, but I see the others are also _anchors_, so don't bother.

public/components/GeoPlot.vue
```diff
@@ -284,6 +294,7 @@ export default Vue.extend({
       isClustering: false,
       isColoringByConfidence: false,
       confidenceIconClass: "confidence-icon",
+      isSatelliteView: false,
```
YohannParis, line 297: to make it easier to read it would be nice to keep that list in an alphabetical order.

public/components/GeoPlot.vue
```diff
@@ -1277,12 +1330,18 @@ export default Vue.extend({
   display: inline;
   position: absolute;
 }
+.map-toggle:hover {
```
YohannParis, line 1333: those `:hover` and `.active` can be factorized. Maybe by creating a `.map-control` class to use on all buttons.

public/components/GeoPlot.vue
```diff
@@ -712,6 +730,28 @@ export default Vue.extend({
     onFocusOut() {
       this.isImageDrilldown = false;
     },
+    mapToggle() {
+      this.isSatelliteView = !this.isSatelliteView;
+      this.map.remove(this.tileRenderer); // remove old tile renderer to destroy the buffers hold the previous tile set
+      this.tileRenderer = new lumo.TileLayer({
```
YohannParis, line 736: From Line736 to 744 can be it's own method and be used in both `mapToggle()` and `createLumoMap()` methods.

YohannParis (approved):
Apart from optional comments, this is good to go.

public/components/GeoPlot.vue
Zac-hills: Cleanup for this component is coming down the pipeline. This will get refactored at some point. The idea is to move all of the util out and move this list of variables into their own objects to make it much cleaner.

public/components/GeoPlot.vue
Zac-hills: Added createMapLayers() to encompass this bit of code. Thanks!

PR #2074, opened 2020-11-14
#2034 - train test split ux & api updates

### LINKED ISSUE #2034: Add train/test split ratio to model create controls
Add a slider to control the train / test split ratio to the model create settings panel.  It should initialize to the system default.

karlolsonuc:
Closes #2034. New UX in settings modal plus supporting store, component and utility updates now send back the mlSplit percentage split to the API/backend while retrieving the the default split from the config thanks to changes to the config.go to pass those fields down in the config grab in the front end with fallbacks and guard rails on both the API and UX code to make sure things leverage default when not otherwise set. Updates to dataset_persist & solution_request to leverage the posted back trainTestSplit field as well, and fall back to the config when that's undefined.

### DIFF
--- a/api/compute/dataset_persist.go
+++ b/api/compute/dataset_persist.go
@@ -363,6 +363,7 @@ type persistedDataParams struct {
 	TaskType           []string
 	TmpDataFolder      string
 	Quality            string
+	TrainTestSplit     float64
 }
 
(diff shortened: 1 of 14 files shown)


Zac-hills (approved):
looks good to me!

PR #2079, opened 2020-11-18
Lowshot learning view

Zac-hills:
Submitting a PR due to the volume of code changes. The label view is at a "bare bones functioning" point.

### Current features
- LowShotLabeling view
- Supports facet highlighting
- Supports data annotation
- Supports data export (with filter params. Allowing us to export specific subset of the data!)

### Unresolved issues
Screenshot from 2020-11-18 14-35-45
- Facet is not displaying the proper labels, something odd is happening there. The correct labels are displayed in the tabular view.

- CSV needs to be downloaded after export. (currently, the data is written to the outputs folder)
- Visual indicator of positive or negative.
- Some sort of way to determine if dataset needs to be cloned
Screenshot from 2020-11-18 14-39-25
- Some layout issues at the top

### DIFF
--- a/api/model/storage/postgres/dataset.go
+++ b/api/model/storage/postgres/dataset.go
@@ -216,7 +216,7 @@ func (s *Storage) parseData(rows pgx.Rows) ([][]string, error) {
 }
 
 // FetchDataset extracts the complete raw data from the database.
-func (s *Storage) FetchDataset(dataset string, storageName string, invert bool, filterParams ...*api.FilterParams) ([][]string, error) {
+func (s *Storage) FetchDataset(dataset string, storageName string, invert bool, filterParams *api.FilterParams) ([][]string, error) {
 	// get data variables (to exclude metadata variables)
 	vars, err := s.metadata.FetchVariables(dataset, true, false)
 	if err != nil {
@@ -225,8 +225,7 @@ func (s *Storage) FetchDataset(dataset string, storageName string, invert bool, 
 	filteredVars := []*model.Variable{}
 	// only include data with distilrole data and index
 	for _, v := range vars {
-		check := v.DistilRole == model.VarDistilRoleIndex || v.DistilRole == model.VarDistilRoleData
-		if check {
+		if model.IsTA2Field(v.DistilRole, v.SelectedRole) {
 			filteredVars = append(filteredVars, v)
 		}
 	}
@@ -234,23 +233,15 @@ func (s *Storage) FetchDataset(dataset string, storageName string, invert bool, 
 	for _, v := range filteredVars {
 		varNames = append(varNames, fmt.Sprintf("COALESCE(\"%s\", '') AS \"%s\"", v.Name, v.Name))
 	}
-
-	if len(filterParams) > 0 {
-		wheres := []string{}
-		paramsFilter := make([]interface{}, 0)
-		wheres, paramsFilter = s.buildFilteredQueryWhere(dataset, wheres, paramsFilter, "", filterParams[0], invert)
-		if len(wheres) > 0 {
-			sql := fmt.Sprintf("SELECT %s FROM %s WHERE %s;", strings.Join(varNames, ", "), getBaseTableName(storageName), strings.Join(wheres, " AND "))
-			res, err := s.client.Query(sql, paramsFilter...)
-			if err != nil {
-				return nil, errors.Wrapf(err, "unable execute query to extract dataset")
-			}
-
-			return s.parseData(res)
-		}
+	wheres := []string{}
+	paramsFilter := make([]interface{}, 0)
+	wheres, paramsFilter = s.buildFilteredQueryWhere(dataset, wheres, paramsFilter, "", filterParams, invert)
+	where := ""
+	if len(wheres) > 0 {
+		where = "WHERE" + strings.Join(wheres, " AND ")
 	}
-	sql := fmt.Sprintf("SELECT %s FROM %s;", strings.Join(varNames, ", "), getBaseTableName(storageName))
-	res, err := s.client.Query(sql)
+	sql := fmt.Sprintf("SELECT %s FROM %s %s;", strings.Join(varNames, ", "), getBaseTableName(storageName), where)
+	res, err := s.client.Query(sql, paramsFilter...)
 	if err != nil {
 		return nil, errors.Wrapf(err, "unable execute query to extract dataset")
 	}
@@ -336,7 +327,7 @@ func (s *Storage) createViewFromMetadataFields(storageName string, fields map[st
 }
 
 // AddVariable adds a new variable to the dataset.
-func (s *Storage) AddVariable(dataset string, storageName string, varName string, varType string, defaultVal ...string) error {
+func (s *Storage) AddVariable(dataset string, storageName string, varName string, varType string, defaultVal string) error {
 	// check to make sure the column doesnt exist already
 	dbFields, err := s.getDatabaseFields(fmt.Sprintf("%s_base", storageName))
 	if err != nil {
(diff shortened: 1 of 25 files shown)


Zac-hills:
Added Yohann upon request.

api/model/storage/postgres/dataset.go
```diff
@@ -216,20 +216,40 @@ func (s *Storage) parseData(rows pgx.Rows) ([][]string, error) {
...
 	vars, err := s.metadata.FetchVariables(dataset, true, false)
 	if err != nil {
 		return nil, err
 	}
-
-	varNames := []string{}
+	filteredVars := []*model.Variable{}
+	// only include data with distilrole data and index
 	for _, v := range vars {
+		check := v.DistilRole == model.VarDistilRoleIndex || v.DistilRole == model.VarDistilRoleData
```
phorne-uncharted, line 228: https://github.com/uncharted-distil/distil-compute/blob/master/model/model.go#L521

Should probably use this function instead

api/model/storage/postgres/dataset.go
```diff
@@ -216,20 +216,40 @@ func (s *Storage) parseData(rows pgx.Rows) ([][]string, error) {
...
+	varNames := []string{}
+	for _, v := range filteredVars {
 		varNames = append(varNames, fmt.Sprintf("COALESCE(\"%s\", '') AS \"%s\"", v.Name, v.Name))
 	}
 
-	sql := fmt.Sprintf("SELECT %s FROM %s;", strings.Join(varNames, ", "), getBaseTableName(storageName))
+	if len(filterParams) > 0 {
+		wheres := []string{}
+		paramsFilter := make([]interface{}, 0)
+		wheres, paramsFilter = s.buildFilteredQueryWhere(dataset, wheres, paramsFilter, "", filterParams[0], invert)
```
phorne-uncharted, line 241: Why accept a list if only the first filter params is used? If this is an attempt at optional parameter, I would suggest against this approach.

I would also suggest that the if block be used to generate a where clause and then have it continue on to the normal execution path (prevents code duplication in this case).

api/model/storage/postgres/dataset.go
```diff
@@ -334,6 +354,9 @@ func (s *Storage) AddVariable(dataset string, storageName string, varName string
 	if !found {
 		// add the empty column to the base table and the explain table
 		sql := fmt.Sprintf("ALTER TABLE %s_base ADD COLUMN \"%s\" TEXT;", storageName, varName)
+		if len(defaultVal) > 0 {
+			sql = fmt.Sprintf("ALTER TABLE %s_base ADD COLUMN \"%s\" TEXT DEFAULT '%s';", storageName, varName, defaultVal[0])
```
phorne-uncharted, line 358: Similar to the above comment, I would suggest against using the ... operator as some type of optional parameter passing approach. I would also suggest the if should set a variable that is used in the sql statement creation to minimize duplication.

public/store/dataset/actions.ts
```diff
@@ -1,4 +1,5 @@
 import axios, { AxiosResponse } from "axios";
+import { data } from "jquery";
```
YohannParis, line 2: I don't see it being used on this file.

public/views/Labeling.vue
```diff
@@ -0,0 +1,212 @@
...
+      dataset: this.dataset,
+    });
+    // pull the cloned data
+    viewActions.updateLabelData(this.$store);
+  },
+});
+</script>
+
+<style scoped>
+.flex-direction-down {
```
karlolsonuc, line 203: Leverage bootstrap flex classes rather than create new classes.

api/model/storage/postgres/dataset.go
```diff
@@ -389,6 +412,10 @@ func (s *Storage) AddField(dataset string, storageName string, varName string, v
 	}
 
 	sql := fmt.Sprintf("ALTER TABLE %s ADD COLUMN \"%s\" %s;", storageName, varName, postgres.MapD3MTypeToPostgresType(varType))
+	if len(defaultVal) > 0 {
+		sql = fmt.Sprintf("ALTER TABLE %s ADD COLUMN \"%s\" %s DEFAULT '%s';", storageName, varName, postgres.MapD3MTypeToPostgresType(varType), defaultVal[0])
```
phorne-uncharted, line 416: Like above, I suggest not using the ...string as a way to pass optional parameters.

public/views/Labeling.vue
```diff
@@ -0,0 +1,212 @@
...
+  flex-direction: column;
+}
+.label-data-slot {
+  height: 80%;
+}
+.facet-wrapper {
```
karlolsonuc, line 209: Leverage bootstrap margin classes rather than create new classes.

api/model/storage/postgres/dataset.go
```diff
@@ -389,6 +412,10 @@ func (s *Storage) AddField(dataset string, storageName string, varName string, v
...
+	}
+	fmt.Println(sql)
```
phorne-uncharted, line 418: Most likely holdover from testing

public/store/dataset/actions.ts
karlolsonuc, line 2: unneeded import?

api/routes/add_field.go
```diff
@@ -0,0 +1,116 @@
...
+		} else {
+			err := dataStorage.AddVariable(dataset, storageName, name, fieldType)
+			if err != nil {
+				handleError(w, err)
+				return
+			}
+		}
+		displayName := name
+		if params["displayName"] != nil {
+			displayName, ok = params["defaultValue"].(string)
```
phorne-uncharted, line 96: I think this should be looking for "displayName" and not "defaultValue"

api/model/storage/postgres/dataset.go
```diff
@@ -216,20 +216,40 @@ func (s *Storage) parseData(rows pgx.Rows) ([][]string, error) {
...
 		varNames = append(varNames, fmt.Sprintf("COALESCE(\"%s\", '') AS \"%s\"", v.Name, v.Name))
 	}
 
-	sql := fmt.Sprintf("SELECT %s FROM %s;", strings.Join(varNames, ", "), getBaseTableName(storageName))
+	if len(filterParams) > 0 {
+		wheres := []string{}
+		paramsFilter := make([]interface{}, 0)
+		wheres, paramsFilter = s.buildFilteredQueryWhere(dataset, wheres, paramsFilter, "", filterParams[0], invert)
+		if len(wheres) > 0 {
+			sql := fmt.Sprintf("SELECT %s FROM %s WHERE %s;", strings.Join(varNames, ", "), getBaseTableName(storageName), strings.Join(wheres, " AND "))
```
karlolsonuc, line 243: maybe assign` strings.Join(varNames, ", ")` to a new variable since we have to use it twice possibly.

api/model/storage/postgres/dataset.go
```diff
@@ -216,20 +216,40 @@ func (s *Storage) parseData(rows pgx.Rows) ([][]string, error) {
...
+			res, err := s.client.Query(sql, paramsFilter...)
+			if err != nil {
+				return nil, errors.Wrapf(err, "unable execute query to extract dataset")
+			}
 
+			return s.parseData(res)
+		}
+	}
+	sql := fmt.Sprintf("SELECT %s FROM %s;", strings.Join(varNames, ", "), getBaseTableName(storageName))
```
karlolsonuc, line 252: second varNames join

karlolsonuc (approved):
Nothing super important, but just double check the comments I left before any merge back.

api/routes/add_field.go
```diff
@@ -0,0 +1,116 @@
...
+			handleError(w, errors.New(errorMsg+"name"))
+			return
+		}
+		fieldType, ok := params["fieldType"].(string)
+		if !ok {
+			handleError(w, errors.New(errorMsg+"fieldType"))
+			return
+		}
+		// update postgres
+		if params["defaultValue"] != nil {
```
phorne-uncharted, line 76: I suggest structuring this code to instead use an if to extract the default value (if present) if possible. This would remove the duplication of the AddVariable call and the need for an else block. The exact implementation may depend on how the defaultValue is passed down to the function.

karlolsonuc:
Looking much better, I'll wait for @phorne-uncharted's ok before merging back.

PR #2081, opened 2020-11-20
Image annotation visual

Zac-hills:
- Added dataset support for datasets other than remote sensing
- Added modal for changing displayName for the label. (defaults to LowShotLabel)
- Added label visualization for images
- Fixed Pagination


karlolsonuc (approved):
Looks Good.

PR #2085, opened 2020-11-24
Image attention

Zac-hills:
- Image attention for the result view
- Currently, works for image-mosaic (tabular needs to be changed to supply the row data)
- Added url query param imageAttention
- Applies filter if imageAttention=true
Peek 2020-11-24 10-41


api/model/storage.go
```diff
@@ -22,6 +22,7 @@ import (
 
 	"github.com/uncharted-distil/distil-compute/metadata"
 	"github.com/uncharted-distil/distil-compute/model"
+	// api "github.com/uncharted-distil/distil/api/model"
```
phorne-uncharted, line 25: Should be deleted

PR #2093, opened 2020-11-26
pagination fix for image mosaic

Zac-hills:


YohannParis (approved):

YohannParis:
You can check with @cdbethune , but I think for things like that you can "cowboy" merge them onto master.

PR #2096, opened 2020-11-27
Data Explorer proof of concept

### LINKED ISSUE #2045: Protoype implementation of data explorer
Start prototyping move to single data explorer screen in line with target design.  It would be best if we could bring it up as a separate view and build in parallel with the existing code base, eventually switching the app over to that view when we feel its ready.  There are still some points in the provided design that need to be refined, but this will at least let us start assessing the impact and provide a platform to test out different approaches.

YohannParis:
Closes #2045

This is a proof of concept for the[ _Data Explorer_ imagined by Pascale](https://docs.google.com/presentation/d/1G42o3o479BXPah855cq5lgEj9PBDYd4zZcxGDGEPg50/edit#slide=id.ga4ab1f9d58_0_6).

I will go through it on Monday meeting to show what I have done, as it can be simpler than going through all the pages of codes.

To access it, click on the logo to activate the new __Prototype__ flag off the _App_ which will turn the logo blue.

Screen Shot 2020-11-27 at 18 28 05

Then in each _Dataset_ preview, a new button gives access to the _Data Explorer_ page:

Screen Shot 2020-11-27 at 18 29 37


karlolsonuc (approved):
Looks good to me, double checked it on my local as well.

public/views/DataExplorer.vue
```diff
@@ -0,0 +1,376 @@
...
+    ActionColumn,
+    AddVariablePane,
+    DataSize,
+    FacetListPane,
+    LeftSidePanel,
+    ImageMosaic,
+    SearchInput,
+    SelectDataTable,
+    SelectGeoPlot,
+    // SelectGraphView,
```
Zac-hills, line 128: This is for later right?

Zac-hills (approved):
looks good

public/views/DataExplorer.vue
YohannParis: Exactly, these are used on the _Select Training_ page, but commented out, so I added them just to not forget about their implementation.

PR #2101, opened 2020-12-03
Timeseries optimization

### LINKED ISSUE #2094: Timeseries performance
UI around large time series datasets like terra-80 and spawn needs an optimization pass.  The scrolling is choppy when timeseries are loading in, memory doesn't appear to be released when pages are changed, and switching between models in the result screen seems to amplify problems.  We need to do a profiling pass to figure out where the issues are, and find opportunities to remove bottlenecks that are causing inconsistent responses  and high memory use.

Zac-hills:
closes #2094 

- Changed the behaviour for fetching time series from a sinflge fetch per time series to a page fetch
- Added unique trails for each tabular view (gives the tabular view complete ownership over the timeseries allowing for deletion without component ownership collision)

Peek 2020-12-03 12-09

### Issues:
- Timeseries tables are not sortable (until we handle sorting in app)
- The timeseries is being stored in a rather redundant fashion requiring multiple Vue.deletes which are quite slow. (Possible revisit and consolidating the timeseries into one dictionary containing the entire timeseries object)


api/model/storage/postgres/timeseries.go
```diff
@@ -262,27 +290,81 @@ func (f *TimeSeriesField) fetchRepresentationTimeSeries(categoryBuckets []*api.B
...
-		"ON timeline.TimeStamps = filteredEvents.TimeStamps ORDER BY timeline.TimeStamps",
-		xColName, storageName, xColName, yColName, storageName, where)
+	query := fmt.Sprintf("SELECT ARRAY_AGG(filteredEvents.TimeStamps ORDER BY filteredEvents.TimeStamps), ARRAY_AGG(COALESCE(filteredEvents.Counts, 'NaN') ORDER BY filteredEvents.TimeStamps), filteredEvents.series_key FROM "+
+		"(SELECT \"%s\" as TimeStamps, \"%s\" as Counts, \"%s\" as series_key FROM %s %s ) filteredEvents "+
+		"GROUP BY filteredEvents.series_key",
+		xColName, yColName, timeseriesColName, storageName, where)
 
 	// execute the postgres query
-	res, err := s.client.Query(query, params...)
+	res, err := s.client.Query(query)
```
phorne-uncharted, line 364: I'm a bit worried that we no longer pass in params to the query. Does the where clause not use any parameters anymore? What about when some filters are applied?

api/model/storage/postgres/timeseries.go
```diff
@@ -262,27 +290,78 @@ func (f *TimeSeriesField) fetchRepresentationTimeSeries(categoryBuckets []*api.B
...
+	return res, nil
+}
+
+func (s *Storage) parseDateTimeSet(rows pgx.Rows) (*map[time.Time]float64, *[]time.Time, error) {
+	result := map[time.Time]float64{}
+	if rows != nil {
+		defer rows.Close()
+	}
+	timeArr := []time.Time{}
+	rows.Next() // should only be 1 row
```
phorne-uncharted, line 308: May want to put in place an if to make sure there's a row. It could lead to a more graceful error.

karlolsonuc (approved):
I notice we have a lot of repetition in the mutators in different views, so I wonder if the underlying logic/patterns for time series updating/removal couldn't be pulled out into a util of some sort. Otherwise though, it looks great.

api/model/storage/postgres/timeseries.go
```diff
@@ -308,35 +330,32 @@ func (s *Storage) FetchTimeseries(dataset string, storageName string, timeseries
...
 		}
 	}
-
-	// sort the timeseries
-	sort.Slice(response, func(i, j int) bool {
-		return response[i].Time < response[j].Time
-	})
-
-	// sum duplicate timestamps
-	response = removeDuplicates(response)
```
cdbethune, line 318: Are the sort and duplicate summing still happening elsewhere?  This is needed for cases where you series key settings don't yield a single timeseries.  Its hardcoded to a sum right now, but should support different aggregations down the road.

api/model/storage/postgres/timeseries.go
```diff
@@ -346,16 +365,17 @@ func (s *Storage) FetchTimeseriesForecast(dataset string, storageName string, ti
...
 		FROM %s data INNER JOIN %s result ON data."%s" = result.index
 		%s
-		ORDER BY %s`,
-		xColName, storageName, s.getResultTable(storageName),
-		model.D3MIndexFieldName, where, xColName)
+		GROUP BY %s`,
+		xColName, timeseriesColName, storageName, s.getResultTable(storageName),
+		model.D3MIndexFieldName, where, timeseriesColName)
 
+	fmt.Println(query)
```
cdbethune: remove?

api/model/storage/postgres/timeseries.go
```diff
@@ -385,19 +405,20 @@ func (s *Storage) FetchTimeseriesForecast(dataset string, storageName string, ti
 			return nil, err
 		}
 	}
-	// Sum duplicate timestamps
-	response = removeDuplicates(response)
```
cdbethune, line 389: Needs to be done for each timeseries.

api/routes/timeseries.go
```diff
@@ -51,6 +53,19 @@ func TimeseriesHandler(metaCtor api.MetadataStorageCtor, ctorStorage api.DataSto
 			handleError(w, errors.Wrap(err, "Unable to parse post parameters"))
 			return
 		}
+		t, ok := params["timeseriesUris"].([]interface{})
+		if !ok {
+			handleError(w, errors.New("Missing timeseriesUris from query"))
+			return
+		}
+		timeseriesUris := []string{}
```
cdbethune, line 57: Acronyms should maintain case - https://github.com/golang/go/wiki/CodeReviewComments#initialisms

I don't think we've done this universally, but we should try to adhere to the standard.  The `params` look up string is tied to the client side JSON, so nothing you can do there.  I think the mixed case for URI has slipped in in a few spots beyond this.

cdbethune (changes requested):

cdbethune (approved):

PR #2115, opened 2020-12-10
Dataset Summary Feature Listing

### LINKED ISSUE #1874: Dataset Listing Feature Count
The dataset listing counts every feature rather than just data features. The count should reflect only the basic features that are part of the dataset, excluding metadata & grouping features.

phorne-uncharted:
Fixes #1874 

Filter features included in dataset summaries to only contain features visible in the details screen. Features that are hidden in groups need to not be included, but the group itself does need to be included.


karlolsonuc (approved):
Looks good to me.

PR #2116, opened 2020-12-10
Lowshot view

Zac-hills:
- LowShot view updated
- Added popup that displays the lowshot results
- Sorts main views by lowshot ranking
- fixed bug where tasks were not updating

### DIFF
--- a/api/task/pipelines.go
+++ b/api/task/pipelines.go
@@ -60,7 +60,7 @@ func SetClient(computeClient *compute.Client) {
 }
 
 func submitPipeline(datasets []string, step *description.FullySpecifiedPipeline) (string, error) {
-	return sr.SubmitPipeline(client, datasets, nil, nil, step)
+	return sr.SubmitPipeline(client, datasets, nil, nil, step, false)
 }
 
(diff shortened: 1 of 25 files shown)


karlolsonuc (approved):
Looks good to me.

api/task/pipelines.go
```diff
@@ -60,7 +60,7 @@ func SetClient(computeClient *compute.Client) {
 }
 
 func submitPipeline(datasets []string, step *description.FullySpecifiedPipeline) (string, error) {
-	return sr.SubmitPipeline(client, datasets, nil, nil, step)
+	return sr.SubmitPipeline(client, datasets, nil, nil, step, false)
```
phorne-uncharted, line 63: I'm worried that no pipelines will be cached now. We definitely want a lot of them to still be cached (PCA, clustering, etc.).

cdbethune: Right - clustering and MI ranking are the two that benefit the most from caching, and can slow down model runs since the ta2 process requests serially.

Zac-hills: that is a whoops. Good catch. I will add the param to submitPipeline and fix that.

Zac-hills: Fixed on the recent push.

cdbethune (approved):

PR #2124, opened 2020-12-11
Lowshot view

### LINKED ISSUE #2118: Support tile selection from map popup
Clicking on a map rectangle launches the drilldown view that displays nearby tiles in correct geosptial relation to one another.  It would be useful to be able to select and mark tiles as pos/neg samples from this view, as a user can browse the streetmap/imagery in the map view to find tiles of interest.

Zac-hills:
- Added color by rank and fixed random suggestions
- Fixed pipeline caching
- Added support for image annotation in map view
closes #2118

### DIFF
--- a/public/util/data.ts
+++ b/public/util/data.ts
@@ -162,7 +162,7 @@ export function getAllDataItems(includedActive: boolean): TableRow[] {
     : datasetGetters.getHighlightedExcludeTableDataItems(store);
   const highlighted = tableData
     ? tableData.map((h) => {
-        return { ...h, isExcluded: true };
+        return { ...h, isExcluded: true }; // adding isExcluded for the geoplot to color it gray
       })
     : [];
   return includedActive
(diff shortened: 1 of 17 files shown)


public/util/data.ts
```diff
@@ -155,6 +155,20 @@ export function parseBinaryScoreResponse(res: BinaryScoreResponse): RankedSet {
...
   return rankedSet;
 }
+// include the highlight
+export function getAllDataItems(includedActive: boolean): TableRow[] {
+  const tableData = includedActive
+    ? datasetGetters.getHighlightedIncludeTableDataItems(store)
+    : datasetGetters.getHighlightedExcludeTableDataItems(store);
+  const highlighted = tableData
+    ? tableData.map((h) => {
+        return { ...h, isExcluded: true };
```
YohannParis, line 165: Could you explain with a comment why you need to add a new property `isExcluded`

YohannParis (approved):
Still waiting to do some monkey testing with that new view, but the code looks straight.

public/util/data.ts
YohannParis: Thank you, that makes things so much clearer.

cdbethune (approved):

Issue #2075, opened 2020-11-16
Add LexBar to application


cdbethune:
Follow on to #2033.  We need to translate our current highlighting functionality to work with the lexbar now that we've got an idea of how it all works, and references to other projects in the company that are using it.  Synch up with @YohannParis to get a dump from him on what he found.  First step will be to plan how we want to go about introducing it into the system, as it is a pretty major change that will probably take a while to implement.

YohannParis:
@karlolsonuc and I had a long talk about what I learn from _LexBar_ and how it can be implemented.
It can be resumed to:
- Beefing up the backend to send more information about _variables_ details
- Creating a wrapper around the external _LexBar_ library in Vue.js
- Keep using our current Store filtering model 
- Creating a utility to translate our current filtering model to _LexBar_ query language

YohannParis:
I have access to _World Modeler_, and its spin-off written in Typescript _ASKE_, codebases; so we do not have to start completely from scratch.

karlolsonuc:
The work can probably be broken up as follows:
1. Three public utility functions in a new util/lex.ts file: One that turns the list of variables into a Lex Bar language; one that turns the highlight/filter from our API into a Lex Bar query; and another that turns a Lex Bar query into our existing highlight/filter. These may have supporting private functions for any reused patterns and helpers, some of which may be similar to the state function wrappers from world modellers.
2. Component wrapper with props for a Lex Bar language & and a lex query, and that has an onChange bind that ties to onQueryChanged events in the Lex Bar. The function passed to onChange bind will leverage the Lex Bar query-to-highlight/filter utility function, then pass the result of that to the existing filter/highlight mutation/update functions.
3. Replacing the existing tag bar in each of the components that call it with the Lex Bar component
4. Any styling we want to show the difference between filters and highlights like we had before in the new Lex Bar.

karlolsonuc:
Possible aside on 2. - we might want to see how much duplication there is in consuming the generic lex bar wrapper. If there's a lot, it could be worth building an intermediate wrapper that does all of the utility calls and takes in the distil formatted information (variables & filters) such that the consuming components and the lex bar component don't have to care about that for maximum reusability without code duplication.

cdbethune:
I think we can close this one off, and track any updates/bugs as new issues.

PR #2126, opened 2020-12-14
#2075 lex bar utils

karlolsonuc:
- added lex.ts with variables to lex language, lex query to filters and filters to lex query utility functions
- factored lex bar language stuff from the searchBar wrapper component for the lex bar to lex.ts so that searchBar wrapper can focus more on display concerns, added needed style imports for date filters, wired up to actually populate filters in the lex bar.
- added bulk filter update to filter.ts utility as lex queries (unlike filter badges) provide use the entire current filter, so it are easier to just fully replace than update per variable. 
- added an event fired function to the data explorer that binds to lex-query event, passes the query to the utility translator, then updates the filter in the route with the new bulk filter utility. 
- factored out dateToNum and numToDate utilities from facetDateTime view to types.ts for reuse as utility functions in

### DIFF
--- a/public/components/facets/FacetDateTime.vue
+++ b/public/components/facets/FacetDateTime.vue
@@ -57,7 +57,7 @@ import {
   facetTypeChangeState,
 } from "../../util/facets";
 import { DATETIME_FILTER } from "../../util/filters";
-import { DATETIME_UNIX_ADJUSTMENT } from "../../util/types";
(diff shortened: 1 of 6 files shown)


YohannParis (approved):
Works great!
It would be nice if the basic string filtering (i.e. _Position_ for _Baseball_ dataset) does an insensitive caps filtering.

PR #2129, opened 2020-12-16
Rank & Confidence Explanation

phorne-uncharted:
Added rank and confidences to the explain_values field to have common querying and storage for explanation output. Rank is not added to the client at all. Once it gets added, it may be beneficial to rework some of the implementation. Some specific code for confidence had to be kept to play nicely with the client, but at some point the client should probably be cleaned up a bit to not require the specific handling.


public/store/results/actions.ts
```diff
@@ -584,6 +584,7 @@ export const actions = {
       solution,
       key,
       label,
+      "summary",
```
karlolsonuc, line 587: Maybe add some string constants for "summary" and "confidence" but it's not a blocker.

karlolsonuc (approved):
Consider my note, but otherwise it's looking good.

public/store/results/actions.ts
phorne-uncharted: Added some constants to cover those two values

PR #2134, opened 2020-12-17
Updated image query to use data endpoint rather than websocket

Zac-hills:
- Changed Image Query data flow by fetching data through an HTTP request rather than websocket
- Added VarDistilRoleSystemData (Responsible for denoting data that is important for the system not the user)
- Changed two key functions on the backend Meta.FetchVariables and Postgres.FetchData
- FetchVariables now can incorporate the new DistilRole
- FetchData can now accept a variable to order by (nil is acceptable for the previous behaviour)
- FetchData orderBy variable does not need to reside within the variables being fetched
- The frontend has support for the new orderBy (recommended way is to add the orderBy query param to the url and fetch the value before fetching the data from the webserver)

### DIFF
--- a/api/compute/solution_request.go
+++ b/api/compute/solution_request.go
@@ -640,7 +640,7 @@ func (s *SolutionRequest) PersistAndDispatch(client *compute.Client, solutionSto
 	s.Filters.AddVariable(model.D3MIndexFieldName)
 
 	// fetch the dataset variables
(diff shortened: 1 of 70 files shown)


phorne-uncharted (approved):

karlolsonuc (approved):
Looks good to me.

Issue #2095, opened 2020-11-27
Pre-featurized Gradcam Support


cdbethune:
Gradcam explanations require image feature data to be unpooled.  This is accomplished in pipelines that don't rely on pre-featurized data by setting `pool_features` to false on the `RemoteSensingPreTrained` primitive, which results in the generation of a 2048 dimensional vector per image, where each element consists of 4x4 floating point array (instead of a scalar value, as in the pooled case).  

The pre-featurization pipeline that runs on dataset ingest (when the `add dataset` function is used in the UI) currently runs a pipeline that consists of the satellite image loader and the remote sensing pre-trained primitive, and writes the vector output to a parquet file.  That output is pooled, and cannot be used to generate Gradcam explanations.  In order to do so, we must allow for the pre-featurization pipelne run at ingest to generate pooled or unpooled vectors, and ensure that the parquet file write will allow for the unpooled features to be written out.  At the TA2 level, we need to ensure that the `ParquetDatasetLoader` can load the unpooled features, and pass them along to the `MLPClassifier` in the format it expects.

PR #2136, opened 2020-12-18
adjusted parquet to open more file headers

venkatkorapaty:
fixes uncharted-distil/distil#2095


api/serialization/parquet.go
```diff
@@ -91,7 +92,7 @@ func (d *Parquet) ReadData(uri string) ([][]string, error) {
 	if err != nil {
 		return nil, errors.Wrap(err, "unable to open parquet file")
 	}
-	defer fr.Close()
+	// defer fr.Close()
```
phorne-uncharted, line 95: drop this comment

api/serialization/parquet.go
```diff
@@ -102,13 +103,35 @@ func (d *Parquet) ReadData(uri string) ([][]string, error) {
 	colCount := pr.SchemaHandler.GetColumnNum()
 	rowCount := pr.GetNumRows()
 	dataByCol := make([][]string, colCount)
+	var rLimit syscall.Rlimit
+	syscErr := syscall.Getrlimit(syscall.RLIMIT_NOFILE, &rLimit)
```
phorne-uncharted, line 107: It may be better to do this on startup and set the value once

api/serialization/parquet.go
```diff
@@ -102,13 +103,35 @@ func (d *Parquet) ReadData(uri string) ([][]string, error) {
...
+				return nil, errors.Wrap(err, "unable to open parquet file reader")
+			}
+		}
 		colRaw, err := d.readColumn(pr, i, rowCount)
 		if err != nil {
 			return nil, err
 		}
 		dataByCol[i] = d.columnToString(colRaw, *pr.SchemaHandler.SchemaElements[i+1].Type)
 	}
+	defer fr.Close()
```
phorne-uncharted, line 133: I think these defers are misplaced. At this point, they can probably just be closed. However, if an error occurs reading the column, then they are never closed. I think these defers should be right after opening them. Although since the previously opened one gets closed explicitly, then the whole thing gets a bit messy.

venkatkorapaty: I will just take out the defers for now, then

api/serialization/parquet.go
```diff
@@ -105,10 +104,27 @@ func (d *Parquet) ReadData(uri string) ([][]string, error) {
 	for i := int64(0); i < colCount; i++ {
 		colRaw, err := d.readColumn(pr, i, rowCount)
 		if err != nil {
-			return nil, err
+			fr.Close()
```
venkatkorapaty, line 107: I think this solution would be more elegant than getting the `ulimit -n` and checking if we reached it or not.

main.go
```diff
@@ -260,7 +260,7 @@ func main() {
 	registerRoute(mux, "/distil/export-results/:produce-request-id", routes.ExportResultHandler(pgSolutionStorageCtor, pgDataStorageCtor, esMetadataStorageCtor))
 	registerRoute(mux, "/distil/model-metrics/:task", routes.ModelMetricsHandler(esMetadataStorageCtor))
 	registerRoute(mux, "/ws", ws.SolutionHandler(solutionClient, esMetadataStorageCtor, pgDataStorageCtor, pgSolutionStorageCtor, esExportedModelStorageCtor))
-	registerRoute(mux, "/distil/image-attention/:dataset/:result-id/:index/:opacity/:color-scale", routes.ImageAttentionHandler(pgSolutionStorageCtor, esMetadataStorageCtor))
+	registerRoute(mux, "/distil/image-attention/:dataset/:resultId/:index", routes.ImageAttentionHandler(pgSolutionStorageCtor, esMetadataStorageCtor))
```
venkatkorapaty, line 263: Uh, I don't really know how these changes got here

PR #2153, opened 2021-01-05
Image attention drill down

### LINKED ISSUE #2112: Add image explanation to drilldown view
If we click on an image to see it in a larger resolution, image explanations are not displayed even if enabled.  They should reflect the current enable/disable state.  A button to toggle on/off directly from the drilldown view would be helpful as well, as the user can then see the source image without the explanation to better understand what it.

Zac-hills:
closes #2112 

- Added filter support for drill down
- Added toggle for filter on drill down support (only on result screen)
- Fixed all the errors from from image drill down
- Supports displaying multiple filters (for the carousel feature)
- Fixed route param issues
- Fixed bug with timesplit always applying
Peek 2021-01-05 14-16


karlolsonuc (approved):
Looks good.

PR #2159, opened 2021-01-06
Added deletion, immutability and clone support for datasets

### LINKED ISSUE #2146: Delete datasets from the client
Now that we have the labelling workflow, we're starting to run into a lot more cases where we've got multiple copies of a dataset.  We should add UI to allow for a database to be deleted.

Zac-hills:
closes #2146 
Peek 2021-01-06 14-11

- Added deletion support for cloned datasets
- Integrated the use of the Immutable tag for datasets
- Integrated the use of Clone tag for datasets
- Added new dataset-deletion route
- Added functions for dropping tables and views
- Changed deletion in ES to sync behaviour
- Two step deletion user interface
- Data that is ingested is Immutable and therefore cannot be deleted or altered
- Cloned datasets are mutable

### DIFF
--- a/api/model/storage/elastic/dataset.go
+++ b/api/model/storage/elastic/dataset.go
@@ -48,8 +48,12 @@ func (s *Storage) CloneDataset(dataset string, datasetNew string, storageNameNew
 	// update the id to match the new info
 	ds.ID = datasetNew
 	ds.StorageName = storageNameNew
+	ds.Name = datasetNew
 	ds.Folder = folderNew
 	ds.Source = metadata.Augmented
+	// cloned datasets CAN be altered
+	ds.Immutable = false
+	ds.Clone = true
 
 	return s.UpdateDataset(ds)
 }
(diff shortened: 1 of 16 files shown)


api/model/storage/elastic/dataset.go
```diff
@@ -48,8 +48,12 @@ func (s *Storage) CloneDataset(dataset string, datasetNew string, storageNameNew
 	// update the id to match the new info
 	ds.ID = datasetNew
 	ds.StorageName = storageNameNew
+	ds.Name = datasetNew
```
phorne-uncharted, line 51: do we want the id and names to be identical?

api/util/file.go
```diff
@@ -202,6 +202,12 @@ func RemoveContents(dir string) error {
 			return err
 		}
 	}
+	if includeParent {
+		err = os.RemoveAll(dir)
+		if err != nil {
+			return err
+		}
+	}
```
phorne-uncharted, line 210: this may be better as the start of the function to short-circuit the iteration

api/routes/clone.go
```diff
@@ -54,7 +54,10 @@ func CloningHandler(metaCtor api.MetadataStorageCtor, dataCtor api.DataStorageCt
 			handleError(w, err)
 			return
 		}
-
+		if ds.Clone {
+			handleError(w, errors.New("Cannot make a clone of a clone"))
+			return
+		}
```
phorne-uncharted, line 60: We never want to clone a clone?

PR #2164, opened 2021-01-07
support added for multiple selections for image query

### LINKED ISSUE #2156: Selection tools for bulk labeling
The current process for selecting an item to label requires the user to click on each individual item
- Add lasso tool
- Add bulk label for page

Zac-hills:
Peek 2021-01-07 13-09

closes #2156 

- Added support for shift click selection for image mosaic and tabular view (only for the image query view, it can be added to other views if needed)
- Added select all button which selects all the items on the current page for image mosaic and tabular view


karlolsonuc (approved):
Looks good.

PR #2167, opened 2021-01-08
Commercial Dataset Support

phorne-uncharted:
The commercial datasets were not being ingested in part due to invisible characters in the header leading to discrepancies between the header row field name and the metadata field name. Invisible characters are now stripped from the header row.

The same datasets also highlighted issues with spaces in the header. These were caused by the introduction of the storage name fields to separate the name used by TA2 and the name used by our storage (ES, PG). When that change was done, `colName` was able to have any character, whereas the new field was limited by PH naming restrictions. As part of this PR, the client was changed to use a variable storage name as key rather than `colName`. `colName` refers to the name of the column in the header row of the file (used by TA2) but the server uses the storage name as variable key. Ideally, we would have a set of structures that handle the client-server I/O and represent the necessary data exchanged, but since none of that exists it was more feasible to update the client.

The relevant issue should not be closed yet. It should only be closed once we confirm that staging works.

### DIFF
--- a/api/compute/dataset_persist.go
+++ b/api/compute/dataset_persist.go
@@ -374,7 +374,7 @@ func CreateBatches(schemaFile string, maxBatchSize int) ([]string, error) {
 	for _, v := range meta.GetMainDataResource().Variables {
 		if v.DistilRole == model.VarDistilRoleGrouping {
 			groupColIndex = v.Index
-		} else if v.StorageName == model.D3MIndexFieldName && groupColIndex == -1 {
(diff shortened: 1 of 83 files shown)


cdbethune (commented):
My one suggestion here, which I understand will be a pain, is to maybe rename `storageName` to something that more clearly indicates its purpose - `id` maybe?  That way it is totally clear everywhere in the code that that is what we would be using to uniquely identify a variable.  Thoughts?

PR #2197, opened 2021-01-18
Highlight the predicted column in the table view.

### LINKED ISSUE #2189: Distinguish predicted values in table view
In the table view, the predicted values need to be made more visually distinct than they are currently are.  Pascale' s suggestion is to have a darker vertical line separate the predictions from the rest of the test data.

YohannParis:
fix #2189 

I'm not sure if that design solution is working? I think this is something to ask Pascale to address. Maybe she can work on a solution or colour scheme to highlight a column.

Screen Shot 2021-01-19 at 13 01 37

### DIFF
--- a/public/components/PredictionsDataTable.vue
+++ b/public/components/PredictionsDataTable.vue
@@ -430,6 +430,5 @@ table tr {
 /* Highlight the predicted column */
 .table td.predicted-value {
   border-right: 2px solid var(--gray-900);
-  border-left: 2px solid var(--gray-900);
 }
 </style>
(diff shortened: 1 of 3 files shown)


cdbethune:
I think if the confidence column is there, its okay to leave the darkened bar out.  We're just trying to visually distinguish the predicted data (that normally just has a superscript) from the ground truth.

cdbethune:
Also, this needs to be implemented in the model results screen as well.

YohannParis:
This is ready to be reviewed again.

PR #2205, opened 2021-01-19
Outlier detection pipeline

### LINKED ISSUE #2142: Pipelines for outlier detection
A primitive for outlier detection is being added under uncharted-distil/distil-primitives#230.  Pipelines should be created similar to clustering for remote sensing that use the this new primitive to perform outlier analysis.  A pipeline for outlier detection on tabular data should be also be added.

venkatkorapaty:
fixes #2142


docker-compose.yml
```diff
@@ -23,19 +23,19 @@ postgres:
 #     - $D3MINPUTDIR:$D3MINPUTDIR
 #     - $D3MOUTPUTDIR:$D3MOUTPUTDIR
 
-distil-auto-ml:
```
cdbethune, line 26: probably checked in by mistake?

api/routes/outlier_detecting.go
```diff
@@ -0,0 +1,60 @@
...
+			return
+		}
+
+		datasetMeta, err := metaStorage.FetchDataset(dataset, false, false, false)
+		if err != nil {
+			handleError(w, err)
+			return
+		}
+
+		outlierData, err := task.OutlierDetection(datasetMeta, variable)
```
cdbethune, line 51: Skipped the err check

cdbethune (changes requested):

docker-compose.yml
venkatkorapaty: yes, reverted it.

api/routes/outlier_detecting.go
venkatkorapaty: included it now.

cdbethune (approved):

PR #2216, opened 2021-01-22
#2172 Sort By Imported

### LINKED ISSUE #2172: Search results should rank datamart sets lower
Searching datasets using "terra" currently returns a mix of internal and datamart results, with available (already imported) terra dataset appearing 4 or 5 positions down the list.  We should adjust the results such that  un-imported datamart datasets are ranked lower than those that are already available.

karlolsonuc:
Fixes #2172 - adds options to sort by the imported state of datasets with supporting functions. ascending order puts imported sets first, descending reverses that, and saved models are grouped with imported sets in the all view.

### DIFF
--- a/public/views/Search.vue
+++ b/public/views/Search.vue
@@ -92,6 +92,14 @@
             <b-dropdown-item-button @click="sortFeaturesDesc">
               <i class="fa fa-sort-numeric-desc"></i> Features
             </b-dropdown-item-button>
+            <b-dropdown-item-button @click="sortImportedAsc">
+              <i class="fa fa-long-arrow-down"></i
+              ><i class="fa fa-file"></i> Imported
+            </b-dropdown-item-button>
+            <b-dropdown-item-button @click="sortImportedDesc">
+              <i class="fa fa-long-arrow-down"></i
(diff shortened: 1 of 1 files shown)


public/views/Search.vue
```diff
@@ -393,6 +409,15 @@ export default Vue.extend({
         : result.dataset.variables.length ?? 0;
     },
 
+    getStorageNameFromResult(result: any) {
```
cdbethune, line 412: `any` shouldn't be used here.  I know it was in the existing code, but given that this file is being modified anyways, we should clean it up.  I was curious to see how much work it would entail, so I did a quick clean up on another branch that I can merge into this one if you want. See 7b523c027e54e03525244754db0200f9ea6a9490.  I dealt with a few linting issues as well.

cdbethune (changes requested):

public/views/Search.vue
karlolsonuc: Yeah, I'll merge that in.

karlolsonuc: Merged your changes back to this branch after double checking them on my local, I think it's ready to merge back main now.

cdbethune (approved):

PR #2221, opened 2021-01-29
Multi-variable timeseries support

### LINKED ISSUE #1421: Convert all numerical vars to timeseries when one is selected

### LINKED ISSUE #1757: Exploding Timeseries made of only Dates and Values (no grouping/series id) results in new extra column
Reproduction Steps

- open sunspots data set
- navigate to variable grouping view for timeseries creation from any facet
- group with year as time column and sunspots as value column
- submit the new grouping
- explode the newly created facet
- note the following new unknown/text type:

Screen Shot 2020-07-14 at 12 29 30 PM

There's probably a back end issue with how explode works in this use case vs. when we have at least grouping / series id column, and it's not an issue with geocoordinates which are always only two columns.

cdbethune:
fixes #1757 fixes #1421

We needed to support situations were we have multiple measures per series in a dataset.  An example would be an extension to the terra dataset, where we have series per sector, and would want to include plant height AND soil moisture measurements for each timestamp.  The result would be a time/height series, and a time/moisture series, for each sector.  This required changing the way we reference timeseries variables - previously we did it using the grouping column name, but this meant that we could only support a single variable per series.  We now reference the variables using the top level variable key.  Other grouped variables (geocoords, images, satellite images) should be handled the same way, but that can be done as follow-on work.  I'm hoping we can stop sending the grouping variables down to the client when it requests the variables from the server, but that will need to some testing, and this PR was already getting big enough.

### DIFF
--- a/api/model/storage/postgres/result.go
+++ b/api/model/storage/postgres/result.go
@@ -889,7 +889,7 @@ func (s *Storage) getAverageWeights(dataset string, storageName string, storageN
 	variables []*model.Variable, whereStatement string, params []interface{}) (map[string]float64, error) {
 	variablesSQL := []string{}
 	for _, v := range variables {
-		if model.IsTA2Field(v.DistilRole, v.SelectedRole) && !model.IsIndexRole(v.SelectedRole) {
+		if model.IsTA2Field(v.DistilRole, v.SelectedRole) && !model.IsIndexRole(v.SelectedRole) && !v.IsGrouping() {
 			variablesSQL = append(variablesSQL, fmt.Sprintf("AVG(weights.\"%s\") as \"%s\"", v.Key, v.Key))
 		}
 	}
--- a/api/model/storage/postgres/result.go
+++ b/api/model/storage/postgres/result.go
@@ -80,11 +80,6 @@ func (s *Storage) getResultTargetVariable(dataset string, targetName string) (*m
 	if err != nil {
 		return nil, errors.Wrap(err, "Unable to get target variable information")
 	}
-	if variable.IsGrouping() && model.IsTimeSeries(variable.Type) {
-		// extract the time series value column
-		tsg := variable.Grouping.(*model.TimeseriesGrouping)
-		return s.metadata.FetchVariable(dataset, tsg.YCol)
-	}
 	return variable, nil
 }
 
@@ -340,7 +335,7 @@ func (s *Storage) PersistResult(dataset string, storageName string, resultURI st
 		}
 		indicesParsed[parsedVal] = true
 
-		dataForInsert := []interface{}{resultURI, parsedVal, targetVariable.Key, records[i][targetIndex]}
+		dataForInsert := []interface{}{resultURI, parsedVal, targetHeaderName, records[i][targetIndex]}
 		explainValues, err := s.parseExplainValues(records[i], confidenceIndex, rankIndex)
 		if err != nil {
 			return err
@@ -426,6 +421,7 @@ func (s *Storage) parseFilteredResults(variables []*model.Variable, rows pgx.Row
 						continue
 					} else {
 						v := getVariableByKey(key, variables)
+						key = v.Key
 						label = v.DisplayName
 						if key == target.Key {
 							typ = target.Type
@@ -1002,6 +998,14 @@ func (s *Storage) getHeaderName(dataset string, key string) (string, error) {
 	variable, err := s.metadata.FetchVariable(dataset, key)
 	if err != nil {
 		return "", errors.Wrap(err, "unable fetch variable for name mapping")
+	}
+
+	if variable.IsGrouping() && model.IsTimeSeries(variable.Grouping.GetType()) {
+		tsg := variable.Grouping.(*model.TimeseriesGrouping)
+		variable, err = s.metadata.FetchVariable(dataset, tsg.YCol)
+		if err != nil {
+			return "", errors.Wrap(err, "unable fetch variable for name mapping")
+		}
 	}
 
 	return variable.HeaderName, nil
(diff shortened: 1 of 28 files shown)


api/model/storage/postgres/result.go
```diff
@@ -340,7 +335,7 @@ func (s *Storage) PersistResult(dataset string, storageName string, resultURI st
 		}
 		indicesParsed[parsedVal] = true
 
-		dataForInsert := []interface{}{resultURI, parsedVal, targetVariable.Key, records[i][targetIndex]}
+		dataForInsert := []interface{}{resultURI, parsedVal, targetHeaderName, records[i][targetIndex]}
```
phorne-uncharted, line 338: I'm a bit worried about the switch to header name here. I am not sure what we use that field for, but I would think it makes more sense to assume the value contains the key rather than the header name.

api/model/storage/postgres/timeseries.go
```diff
@@ -781,6 +803,25 @@ func (f *TimeSeriesField) parseHistogram(rows pgx.Rows, mode api.SummaryMode) (*
...
+		Buckets: []*api.Bucket{
+			{
+				Key:   f.Key,
+				Count: 1,
+			},
+		},
+		Extrema: &api.Extrema{
+			Min: float64(1),
+			Max: float64(1),
+		},
```
phorne-uncharted, line 820: In this case, are the extrema supposed to be for the count, or for the underlying values?

api/routes/grouping.go
```diff
@@ -203,19 +215,17 @@ func createGrouping(dataset string, storageName string, groupingType string, raw
...
+			return errors.Wrapf(err, "unable to create new variable %s", tsg.IDCol)
 		}
 
+		// Set the name of the expected cluster column - it doesn't necessarily exist.
+		tsg.ClusterCol = model.ClusterVarPrefix + tsg.IDCol
+
 		// Create a new grouped variable for the time series.
 		groupingVarName := strings.Join([]string{tsg.XCol, tsg.YCol}, task.DefaultSeparator)
-		err = meta.AddGroupedVariable(dataset, groupingVarName, tsg.YCol, model.TimeSeriesType, model.VarDistilRoleGrouping, tsg)
+		err = meta.AddGroupedVariable(dataset, groupingVarName, tsg.YCol, model.TimeSeriesType, model.VarDistilRoleData, tsg)
```
phorne-uncharted, line 228: Why is this now a data variable and not a grouping variable?

api/routes/residuals_summary.go
```diff
@@ -26,6 +26,7 @@ import (
 )
 
 // ResidualsSummary contains a fetch result histogram.
+// CDB: Does this extra level need to be here?
```
phorne-uncharted, line 29: Almost surely not

api/model/storage/postgres/timeseries.go
cdbethune: I think I actually need to go in and make sure that's being calculated properly - that was a placeholder that I meant to upated.

api/routes/grouping.go
cdbethune: The `VarDistilRoleGrouping` role should be applied to a column that is used to group in the case that has to be done.  It shouldn't have been applied to the top level variable.

api/model/storage/postgres/result.go
cdbethune: I think that's a just a result of me having to merge in the `Key` rename.  Everything works, but I'll replace it with `targetVariable.Key` since that's what gets used almost everywhere else.

api/routes/residuals_summary.go
cdbethune: Unused, removing.

api/model/storage/postgres/result.go
cdbethune: No, not a result of the refactor, and this is a legit change.  `targetHeaderName` is generated through a function `getHeaderName` that will return `targetVariable.Key` for a normal variable, or will chase down the appropriate target column in the raw data for grouped variables.  The current example of this is timeseries - the target is the top compound timeseries variable, but the actual value that is predicted at the TA2 level is the the YCol (value) field of the grouped variable.

api/model/storage/postgres/timeseries.go
cdbethune: Looks like this is correct - because it is limited to the single timeseries case (ie. sunspots) there is only every going to be one series, with a min and max of one for its bucket count.

phorne-uncharted (approved):

public/views/VariableGrouping.vue
```diff
@@ -503,31 +515,70 @@ export default Vue.extend({
...
         subIds: ids,
-        hidden: hiddenCols,
+        hidden: hidden,
       };
 
+      const groupings = [] as Grouping[];
+
       if (this.isTimeseries) {
-        const tsGrouping = grouping as TimeseriesGrouping;
+        const tsGrouping = _.cloneDeep(grouping) as TimeseriesGrouping;
```
karlolsonuc, line 532: could we get away with just a [ ...grouping ] spread here or was it not deep enough?

public/views/VariableGrouping.vue
```diff
@@ -503,31 +515,70 @@ export default Vue.extend({
...
+            .filter((v) => Y_COL_TYPES[v.colType])
+            .filter((v) => !this.isIDCol(v.colName))
+            .filter((v) => !this.isXCol(v.colName))
+            .filter((v) => !this.isYCol(v.colName))
+            .filter((v) => !this.isGroupingCol(v))
+            .map((v) => v.colName)
+            .forEach((v) => {
+              // create a new grouping entry for each value variable and add it to
+              // the list to create
+              const tsGrouping = _.cloneDeep(grouping) as TimeseriesGrouping;
```
karlolsonuc, line 554: same question as before on using cloneDeep vs. [...grouping]

karlolsonuc (approved):
Couple of minor questions on using cloneDeep vs. [... grouping] spreads, but otherwise, I think it looks very good.

karlolsonuc (approved):
Excellent.

PR #2223, opened 2021-01-29
2192 Outlier UI

### LINKED ISSUE #2192: Add outlier analytic to UI
We're adding server support for finding outliers in data (see #2142).  This will be a new feature setup similar to clusters, but rather than being a categorical "pattern" value, it will be a new boolean feature where True indicates a sample is an outlier, and False flags otherwise.  The integration into the UI should be similar - there should be a new button added to the analytic bar on the right (where we currently have the cluster and feature ranking buttons), and upon hitting "apply", the outlier feature should be added to the dataset.  The outlier boolean column should appear in the table data, and a facet should be made available to allow for filtering/highlighting.

YohannParis:
closes #2192 

I copied the clustering feature roughly.


api/routes/outlier_detecting.go
```diff
@@ -40,22 +46,128 @@ func OutlierDetectionHandler(metaCtor api.MetadataStorageCtor) func(http.Respons
...
 		}
 
+		// create a result
+		result := OutlierResult{
+			OutlierField: "false",
+		}
+
+		if outlierData != nil {
+			result = OutlierResult{
+				OutlierField: "_outlier",
```
phorne-uncharted, line 70: The name of the field should probably be a constant.

api/routes/outlier_detecting.go
```diff
@@ -40,22 +46,128 @@ func OutlierDetectionHandler(metaCtor api.MetadataStorageCtor) func(http.Respons
...
+		// find the outliers in the dataset
 		outlierData, err := task.OutlierDetection(datasetMeta, variable)
 		if err != nil {
 			handleError(w, err)
 			return
 		}
 
+		// create a result
+		result := OutlierResult{
+			OutlierField: "false",
```
phorne-uncharted, line 65: I haven't looked at the client, but does it check if the variable is named `false` to determine if it worked? It may be better to have a second field in the response that is boolean to determine if the call succeeded or not.

api/routes/outlier_detecting.go
```diff
@@ -21,16 +21,28 @@ import (
...
+type OutlierResult struct {
+	OutlierSuccess bool   `json:"success"`
+	OutlierField   string `json:"outlier"`
+}
+
 // OutlierDetectionHandler generates a route handler that enables outlier detection
-// for either remote sensing or tabular data
+// for either remote sensing or tabular data.
+// Return the name of the variable if the detection has run successfully.
 func OutlierDetectionHandler(metaCtor api.MetadataStorageCtor) func(http.ResponseWriter, *http.Request) {
```
cdbethune, line 43: Clustering and feature ranking both run as a single call, creating columns etc. as part of the task before returning.  Was there a specific reason why outliers have been set up with two distinct routes?

public/components/StatusPanel.vue
```diff
@@ -272,10 +292,12 @@ export default Vue.extend({
         });
 
       // serialize the modes map into a string and add to the route
+      // and update to know that the clustering has been applied.
       const varModesStr = varModesToString(varModesMap);
       const entry = overlayRouteEntry(this.$route, {
         varModes: varModesStr,
         dataMode: DataMode.Cluster,
+        clustering: "1",
```
cdbethune, line 300: Should this be defined as an enum?

public/util/types.ts
```diff
@@ -375,6 +375,10 @@ export function isImageType(type: string): boolean {
   return IMAGE_TYPES.indexOf(type) !== -1;
 }
 
+export function isRemoteSensingType(type: string): boolean {
```
cdbethune, line 378: We're trying to get rid of "remote sensing" as a term in the client, since the actual var type is multiband image.  I would rename this to `isMultibandImageType`.

api/routes/outlier_detecting.go
YohannParis: _Clustering_ is [re-applying the changes at each refresh of the page where the status bar is loaded.](https://github.com/uncharted-distil/distil/pull/2223/files#diff-7799ca076839513ade3b74fdf77f02549db37b0c0dbd82970256ceb5cdc75146R82)
And _Variable Ranking_ [keeps the results in the store and refresh them at every fetch of training variables.](https://github.com/uncharted-distil/distil/pull/2223/files?file-filters%5B%5D=.go&file-filters%5B%5D=.ts#diff-9f14e1174edfb8427cdc67df894afcd7b19e4046f7f42d55be7da91e38ae43baR455)

Because the requirement for this issue was to display the _Outlier Detection_ as a variable, I thought it more straightforward to add it to the back-end datasets; instead of testing on each view or state of the route when to add the outlier results. Now, once the page load, the _Outlier Detection_ runs, and if it's correct, the user can *apply* its result, which adds it to the back-end datasets (SQL & ES.)

public/components/StatusPanel.vue
YohannParis: It is possible, maybe we can refactor that when we redo the route in our blue sky project.

The current choice was made quickly to avoid refresh of the _Clustering_ in a [previous issue.](https://github.com/uncharted-distil/distil/pull/2223/files?file-filters%5B%5D=.go&file-filters%5B%5D=.ts#diff-289d9418bc0c4feaf33efa893939615725e09ae0fc202000f318f43eb8ffcc0eR197)

PR #2275, opened 2021-02-12
2201 remote sensing

### LINKED ISSUE #2201: Unlabelled remote sensing / image archive ingest
Remote sensing or image zip datasets won't import into the application if the images are not stored in subfodlers (that act as label names).

Zac-hills:
- Added support for remote sensing datasets without sub folders
- Changed the way geobounds facet buckets are calculated to avoid IOOR
- Added check to not load the upscale models if shouldScaleImages is false
- Added shouldScaleImages to frontend to disable the button
- 
image

- Note: possible issue with selectTarget view. Currently there is no way to check if there is an available target (and if so we should probably route to the labeling view instead of target). The provided image shows that time stamp can be selected as the target; however, upon selecting timestamp the server errors with unable to find intended model. 
closes #2201

### DIFF
--- a/api/dataset/satellite.go
+++ b/api/dataset/satellite.go
@@ -150,7 +150,7 @@ func (s *Satellite) CreateDataset(rootDataPath string, datasetName string, confi
 	}
 	labelHeader := "label"
 	expectedHeaders := []string{model.D3MIndexFieldName, "image_file", "group_id", "band", "timestamp", "coordinates", labelHeader, "geo_coordinates"}
-	headerNames := expectedHeaders
+	headerNames := append([]string(nil),expectedHeaders...)
 	if len(imageFolders) == 0 {
 		// search just in case the above order is changed at some point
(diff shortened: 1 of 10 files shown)


api/dataset/satellite.go
```diff
@@ -216,8 +235,10 @@ func (s *Satellite) CreateDataset(rootDataPath string, datasetName string, confi
 					d3mIDRunning = d3mIDRunning + 1
 					d3mIDs[groupID] = d3mID
 				}
+				// remove values that are not needed based on the headerNames (expects values, expectedHeaders and headerNames to be IN ORDER)
+				csvLine := removeMissingValues(expectedHeaders, headerNames, []string{fmt.Sprintf("%d", d3mID), path.Base(targetImageFilename), groupID, band, timestamp, coordinates.String(), label, coordinates.ToGeometryString()})
```
phorne-uncharted, line 239: given that every row will have the same transform applied, there should be a way to not duplicate effort quite so much

PR #2285, opened 2021-02-18
images are performing much better

### LINKED ISSUE #2283: Image performance
- The UI is dragging its feet when rendering images

Zac-hills:
Peek 2021-02-18 15-07
The v-observe-visibility was only calling the callback when the element was visible. 
When paging, all the items that had previously been once visible would try to fetch images. 
Potentially 100 request on page. This was the cause of the sluggishness.

closes #2283


karlolsonuc (approved):
Excellent catch. I'll merge it back shortly.

PR #2288, opened 2021-02-19
fixed geoplot not loading images. Fixed timestamp facet.

### LINKED ISSUE #2264: Unable to select table rows
When you click on a table row, the row should highlight in pink, and the facets should update their visuals such that the bars containing the selected row values are likewise highlighted in pink.  This is no longer working.

Zac-hills:
closes #2264
image

- After tracking down the source of the issue:
- https://github.com/uncharted-distil/distil/blob/main/api/model/storage/elastic/variable.go#L429
- The d3mIndex variable had a distilRole of Data causing it to be filtered out. The d3mIndex is required for rowSelection. Upon reingesting the baseball ds shown above. The d3mIndex is given the proper distilRole. In order to fix this we just need to do a reingest of all the datasets.
- Fixed map not loading images
- Fixed timestamp not showing rowSelection (there was a key mismatch)

### DIFF
--- a/public/components/GeoPlot.vue
+++ b/public/components/GeoPlot.vue
@@ -599,7 +599,6 @@ export default Vue.extend({
 
     // Return name of column used as grouping column for the table data
     multibandImageGroupColumn(): string {
-      console.log(datasetGetters.getVariables(this.$store));
       const groupColumns = datasetGetters
         .getVariables(this.$store)
         .find((v) => v.colType === MULTIBAND_IMAGE_TYPE);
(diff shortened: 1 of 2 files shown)


public/components/GeoPlot.vue
```diff
@@ -602,14 +599,11 @@ export default Vue.extend({
 
     // Return name of column used as grouping column for the table data
     multibandImageGroupColumn(): string {
+      console.log(datasetGetters.getVariables(this.$store));
```
phorne-uncharted: Probably want to remove this

PR #2289, opened 2021-02-19
The v-observe has issues where it does not register quick scrolls.

### LINKED ISSUE #2287: Mosaic / table tiles not always loading
To reproduce:

1. Load locust_small dataset
2. Launch labelling view
3. Scroll to bottom.
4. Click on a page further out, scroll top.  Tiles don't ever show up:

image

Zac-hills:
closes #2287
- v-observe-visibility has issues where it wont register quick scrolls. Causing the image preview not to fetch images because it thinks it is not in the viewport.(this unfortunately causes performance issues with a max of 100 fetches on page)

### DIFF
--- a/public/components/ImagePreview.vue
+++ b/public/components/ImagePreview.vue
@@ -111,7 +111,7 @@ export default Vue.extend({
       if (newUrl === null) {
         return;
       }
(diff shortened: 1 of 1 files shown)


karlolsonuc (commented):
Looks okay.

PR #2300, opened 2021-02-22
fixes for intersect vs within

### LINKED ISSUE #2294: Clicking on map image not always resulting in tile view
Clicking on a tile on the map often (but not always) results in a empty tile view displaying.  In the image below, the map tile pointed to by the red arrow was clicked, but the displayed tile view was empty:

image

The console in the client contains errors along the lines of:

```
vue.runtime.esm.js:619 [Vue warn]: Error in mounted hook: "TypeError: Cannot read property '7' of undefined"

found in

---> <DrillDown> at public/components/DrillDown.vue
       <GeoPlot> at public/components/GeoPlot.vue
         <LabelGeoPlot> at public/components/labelingComponents/LabelGeoplot.vue
           <LabelingDataSlot> at public/components/labelingComponents/LabelingDataSlot.vue
             <LabelingView> at public/views/Labeling.vue
               <VariableGrouping> at public/views/VariableGrouping.vue
                 <Root>
```

Zac-hills:
closes #2294

Recently the geo bounds query has changed from ST_WITHIN to ST_INTERSECTS.
This change dealt with buckets not filling correctly.
The issue with the map is it still assumed ST_WITHIN
- Changed the drilldown to just drop any tiles supplied that are not within the supplied bounds (more future proof depending on if the underlying query changes)

### DIFF
--- a/public/components/DrillDown.vue
+++ b/public/components/DrillDown.vue
@@ -183,6 +183,15 @@ export default Vue.extend({
           t.coordinates as LatLngBoundsLiteral
         ).getCenter();
         const indices = this.getIndex(center.lng, center.lat);
+        if (
+          indices.x < 0 ||
+          indices.y < 0 ||
+          indices.x >= this.cols ||
+          indices.y >= this.rows
+        ) {
+          // tile outside defined area
+          return;
+        }
         const invertY = this.rows - 1 - indices.y;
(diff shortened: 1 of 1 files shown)


karlolsonuc (approved):
Looks good!

PR #2311, opened 2021-02-24
validte ulimit on start up

### LINKED ISSUE #1947: Validate ulimit on startup
Parquet files need one reader per-column, and with pre-featurized remote sensing data, we currently work with datasets with ~2048 columns.  This can cause a problem if the file limits for a given systems are not set sufficiently high.  We can check for this setting being too low and correct it at runtime using the same approach as the following:

```golang
func main() {
    var rLimit syscall.Rlimit
    err := syscall.Getrlimit(syscall.RLIMIT_NOFILE, &rLimit)
    if err != nil {
        fmt.Println("Error Getting Rlimit ", err)
    }
    fmt.Println(rLimit)
    rLimit.Max = 999999
    rLimit.Cur = 999999
    err = syscall.Setrlimit(syscall.RLIMIT_NOFILE, &rLimit)
    if err != nil {
        fmt.Println("Error Setting Rlimit ", err)
    }
    err = syscall.Getrlimit(syscall.RLIMIT_NOFILE, &rLimit)
    if err != nil {
        fmt.Println("Error Getting Rlimit ", err)
    }
    fmt.Println("Rlimit Final", rLimit)
}
```

venkatkorapaty:
fixes #1947

### DIFF
--- a/main.go
+++ b/main.go
@@ -69,7 +69,6 @@ func registerRoutePost(mux *goji.Mux, pattern string, handler func(http.Response
 }
 
 func validateULimit(config env.Config) {
-	log.Infof("version: %s built: %s", version, timestamp)
 	var rLimit syscall.Rlimit
 
 	err := syscall.Getrlimit(syscall.RLIMIT_NOFILE, &rLimit)
@@ -105,8 +104,10 @@ func main() {
 		os.Exit(1)
 	}
 
+	log.Infof("Validating ulimit..")
 	validateULimit(config)
 
+	log.Infof("version: %s built: %s", version, timestamp)
 	servicesToWait := make(map[string]service.Heartbeat)
 
 	userAgent := fmt.Sprintf("uncharted-distil-%s-%s", version, timestamp)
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -68,20 +68,51 @@ func registerRoutePost(mux *goji.Mux, pattern string, handler func(http.Response
 	mux.HandleFunc(pat.Post(pattern), handler)
 }
 
-func main() {
+func validateULimit(config env.Config) {
 	log.Infof("version: %s built: %s", version, timestamp)
```
phorne-uncharted, line 72: This log message should stay in the main function

main.go
```diff
@@ -215,6 +246,8 @@ func main() {
 
 	// Ingest the data specified by the environment
 	if config.InitialDataset != "" && !config.SkipIngest {
+		log.Infof("Validating ulimit..")
```
phorne-uncharted, line 249: This log message should be in the validating ulimit function

PR #2320, opened 2021-02-26
Timeseries Creation

phorne-uncharted:
It looks like at some point the d3m index stopped being returned by default from the data fetch. Or some related change was made. Either way, the d3m index field was added in querying for grouping values when building a timeseries (really any component variable, but currently only called by timeseries).


YohannParis (approved):
Works great!

PR #2332, opened 2021-03-03
Render primitives

### LINKED ISSUE #2323: Point maps not working
When a lat/lon type was created for `ACLED_clean` and added to the feature set, the map view was unable to load.  The following exception is thrown by the client:

```
vue.runtime.esm.js:1888 TypeError: Cannot read property 'coordinates' of undefined
    at eval (GeoPlot.vue:1061)
    at Array.forEach (<anonymous>)
    at VueComponent.areaToPoints (GeoPlot.vue:1060)
    at Object.quads (GeoPlot.vue:720)
    at VueComponent.createLumoMap (GeoPlot.vue:800)
    at VueComponent.mounted (GeoPlot.vue:780)
    at invokeWithErrorHandling (vue.runtime.esm.js:1854)
    at callHook (vue.runtime.esm.js:4219)
    at Object.insert (vue.runtime.esm.js:3139)
    at invokeInsertHook (vue.runtime.esm.js:6346)
```

Note that the above exception is just a symptom of a larger problem - it looks like the geoplot is expecting its data to be quads, whereas it use to support point or quad data.  There also seems to be a large amount of dead code present in the file that should be cleaned up.

Zac-hills:
Fix for geo coordinates not working.
closes #2323

### DIFF
--- a/public/components/GeoPlot.vue
+++ b/public/components/GeoPlot.vue
@@ -93,15 +93,6 @@
         />
       </div>
     </b-toast>
-    <button
-      v-show="showExit"
-      type="button"
-      class="close selection-exit"
-      aria-label="Close"
-      :style="exitStyle"
-    >
-      <span aria-hidden="true">&times;</span>
-    </button>
   </div>
 </template>
 
@@ -130,7 +121,6 @@ import {
   RowSelection,
   GeoCoordinateGrouping,
   VariableSummary,
-  MultiBandImageGrouping,
   GeoBoundsGrouping,
 } from "../store/dataset/index";
 import { updateHighlight, highlightsExist } from "../util/highlights";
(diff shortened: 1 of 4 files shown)


public/components/GeoPlot.vue
```diff
@@ -838,16 +773,18 @@ export default Vue.extend({
     getInterestBounds(area: Area): LatLngBoundsLiteral {
       const xDistance = (this.drillDownState.numCols - 1) / 2;
       const yDistance = (this.drillDownState.numRows - 1) / 2;
-      const tileWidth = area.coordinates[1][1] - area.coordinates[0][1];
-      const tileHeight = area.coordinates[1][0] - area.coordinates[0][0];
+      const tileWidth =
+        area.info.coordinates[1][1] - area.info.coordinates[0][1];
```
YohannParis, line 777: I see that you need the tile width and height here and in the _drilldown_ component. 
Maybe you could add it to the class `CoordinateInfo`, so the next reader doesn't need to check if `coordinates[0]` is _Lat_ or _Long_.
You could also add `lat` and `long` as properties for a more straightforward use; thus, only `util/rendering/coordinates.ts` will manage which one is which.

To resume, something like this:
```js
export abstract class CoordinateInfo {
  coordinates: LatLngBoundsLiteral;
  lat: number,
  long: number,
  height: number,
  width: number,
  ...
}
```

YohannParis (approved):
I love the tidying up of the `GeoPlot` component by creating a new `util`.

Everything works as expected, and displays where it should be.

PR #2339, opened 2021-03-04
fixed image labels

### LINKED ISSUE #2326: Correct/incorrect colouring no longer applied to image predictions
Image predictions that are correct should display the single correct abbreviation with a green background.  Incorrect predictions should display the incorrect prediction abbreviation with a red background, and the ground truth abbreviation with the standard background.  We are currently showing no abberviation when correct, and the incorrect abbreviation (with no ground truth) when a prediction is wrong.

image

Zac-hills:
closes #2326 
- Fixed issue with the predicted labels not showing up
- Fixed image labels to pull from result store (I noticed an issue where if one hit refresh the labels would go away due to it pulling from the dataset store)
Screenshot from 2021-03-04 12-11-05

### DIFF
--- a/public/components/ImageLabel.vue
+++ b/public/components/ImageLabel.vue
@@ -57,15 +57,20 @@ export default Vue.extend({
       type: Boolean as () => boolean,
       default: false,
     },
+    isResult: { type: Boolean as () => boolean, default: false },
   },
 
   computed: {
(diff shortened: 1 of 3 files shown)


karlolsonuc (approved):
Looks Good.

PR #2341, opened 2021-03-05
Simplify Dataset Creation From Labelling

phorne-uncharted:
Added functions to update and filter datasets stored on disk. These new functions provide the needed functionality in a more streamlined approach which should be reusable when updating datasets on disk for prefiltering. The code itself is much easier to read and is more intuitive, so should be easier to maintain if we need to expand on it going forward.

As part of these changes, the immutable field of variables is now being used. On ingest, variables are made immutable. When adding a field via the REST API, it is made mutable. Only mutable variables get updated when creating a dataset from user driven labelling.


Zac-hills (approved):
Looks good, much needed addition!

PR #2354, opened 2021-03-10
Map fix

### LINKED ISSUE #2340: Tiles grayed out on zoom display
When in zoom display mode on the map, the tiles are showing up in grayscale.  This should only happen if the tile is currently being filtered as the result of a highlight action.

image

Zac-hills:
- First pass 
- There is a lot of issues surrounding multiple highlights and the way the data was being queried
- This PR fixes issues:
1. With multiple highlight there was no way to remove a geo-coordinate highligh(fix add lex bar to result and prediction view)
2. Fix for lexbar highlight vs highlights
3. The invert does a collective NOT() which converts the sequence of ORs into an and (The desired behaviour is keeping the highlights always as an OR)
4. Fixed the map to support multiple highlights
5. Fixed issue in result screen where the backend always assumes rank is available on the result (causing the route to crash and not provide any other result summary like confidence)
closes #2340

### DIFF
--- a/public/components/GeoPlot.vue
+++ b/public/components/GeoPlot.vue
@@ -712,19 +712,25 @@ export default Vue.extend({
       ) {
         return;
       }
+      const tileMap = new Map<string, Area>();
       const innerArea = this.tableDataToAreas(
         this.areaOfInterestItems.inner
       ) as any[];
       innerArea.forEach((i) => {
         i.gray = 0;
+        tileMap.set(i.imageUrl, i);
       });
       const outerArea = this.tableDataToAreas(
         this.areaOfInterestItems.outer
       ) as any[];
       outerArea.forEach((i) => {
         i.gray = 100;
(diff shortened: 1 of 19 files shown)


public/components/GeoPlot.vue
```diff
@@ -712,19 +712,25 @@ export default Vue.extend({
...
         this.areaOfInterestItems.outer
       ) as any[];
       outerArea.forEach((i) => {
         i.gray = 100;
+        if (!tileMap.has(i.imageUrl)) {
+          tileMap.set(i.imageUrl, i);
+        }
       });
-      this.drillDownState.tiles = innerArea.concat(outerArea);
+      console.log(tileMap);
```
karlolsonuc, line 732: console log to remove.

karlolsonuc (approved):
Looks good, just pull that extra console log and it's ready to merge back, great work.

PR #2377, opened 2021-03-16
join view working again

Zac-hills:


public/util/join.ts
```diff
@@ -60,3 +61,14 @@ export function loadJoinView(
   });
   router.push(entry).catch((err) => console.warn(err));
 }
+
+export function getVariableSummaries(context): VariableSummary[] {
```
cdbethune, line 65: Missing some type info: 

1. `context` arg should be a `Store<DistilState>`
2. `result = []` should be declared as `result = [] as VariableSummary[]` otherwise subsequent use treats it as `any`.

YohannParis (approved):

YohannParis:
Thank you for cleaning that up.

PR #2381, opened 2021-03-17
Fix ImageOreview

### LINKED ISSUE #2365: Media Files and ImagePreview
image

Shown in the above picture the facets are displaying the correct image but the data view is not.
Might be an issue where the ResultDataSlot is not passing in the correct flag to the ImagePreview

YohannParis:
fix #2365 

The `unique-trail` prop was wrongly used to fetch images of type `IMAGE_TYPE`.

The Image file creation is so overblown, why do we save HTMLImageElement when we only need them for their URLs?


public/components/ImagePreview.vue
```diff
@@ -205,56 +134,76 @@ export default Vue.extend({
...
     imageAttentionIsLoaded(): boolean {
       return (
         !!this.solutionId &&
         !!this.row &&
         !!this.files[this.solutionId + this.row.d3mIndex]
       );
     },
+
     image(): HTMLImageElement {
+      if (this.type === IMAGE_TYPE) {
```
YohannParis, line 179: Second line that solves the issue.

Zac-hills (approved):

PR #2383, opened 2021-03-18
Join dataset

### LINKED ISSUE #2357: Join screen missing facets
Facets are not displayed on the join screen - those from each dataset should be available.

image

Zac-hills:
closes #2357
- Added lexbar to join screen (adding support for multiple dataset filtering)
- Fixed variable facets to highlight correctly

### DIFF
--- a/public/components/JoinDataSlot.vue
+++ b/public/components/JoinDataSlot.vue
@@ -17,15 +17,12 @@
 
 <template>
   <div class="join-data-slot d-flex flex-column">
-    <div class="fake-search-input">
-      <div class="filter-badges">
-        <filter-badge
-          v-for="(highlight, index) in activeHighlights"
-          :key="index"
-          :filter="highlight"
-        />
-      </div>
-    </div>
+    <search-bar
+      class="mb-3"
+      :variables="variables"
+      :highlights="routeHighlight"
+      @lex-query="updateFilterAndHighlightFromLexQuery"
+    />
 
     <div class="join-data-container flex-1">
       <div v-if="!hasData" class="join-data-no-results">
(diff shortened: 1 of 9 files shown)


public/views/JoinDatasets.vue
```diff
@@ -29,15 +29,28 @@
...
-      <variable-facets
-        class="col-12 col-md-3 d-flex h-50 pt-2"
-        enable-search
-        enable-type-change
-        enable-highlighting
-        :instance-name="instanceName"
-        :rows-per-page="numRowsPerPage"
-        :summaries="variableSummaries"
-      />
+      <div class="h-100">
```
YohannParis, line 32: You need to add the class `col-md-3` to work on with the next `<div>` on screen smaller than 1000px or so.

YohannParis (changes requested):
- [ ] Deleting existing highlight, will remove some from the other dataset.
- [ ] highlighting a common feature/value in both dataset (i.e. Country/Albania) creates an highlight in both datasets.
- [ ] The variables facets search bars don't work.

public/components/JoinDataSlot.vue
```diff
@@ -56,10 +56,9 @@ import {
   Highlight,
   Variable,
 } from "../store/dataset/index";
-import { createFiltersFromHighlights } from "../util/highlights";
+import { createFiltersFromHighlights, UPDATE_ALL } from "../util/highlights";
 import { Filter, INCLUDE_FILTER } from "../util/filters";
-import { getVariableSummaries } from "../util/join";
-import { updateHighlight, UPDATE_ALL } from "../util/highlights";
+import { updateHighlight } from "../util/highlights";
```
YohannParis, line 61: I don't think we need to separate imports here.

YohannParis (changes requested):

PR #2396, opened 2021-03-22
facet color fix

### LINKED ISSUE #2348: Geocoordinate facet and multiple highlights issue
- The geocoordinate facet seems to break when multiple highlights are involved.
- Notice how dark the color gets after the second highlight
Peek 2021-03-08 09-32

Zac-hills:
Changed all the domain functions to use the baseline buckets for min/max val. This change ensures the boxes are always the same color regardless of filter/highlight

closes #2348


YohannParis (approved):
That makes so much more sense!

PR #2415, opened 2021-03-25
Image rework

### LINKED ISSUE #2398: Optimize image fetch
Implement batched fetching of images similar to what was done for fetching timeseries.

Zac-hills:
closes #2398
- Bulk fetching images has been added
- image-pack route returns an array of image bytes and the imageID for each image
- IMAGE_THREAD_POOL is the env var that depicts how many threads should be used for fetching images (tested 2-6 left it defaulting to 2 for mac books; however, I have not tested more threads on the mac)
- Each DataTable and imageMosiac is responsible for fetching the needed images now (effectively reducing the number of request per page from 100 to 1)
- Removed some debounce from imagePreview it is no longer needed
- Removed element appending in ImagePreview it now has images in the html and changes the src
- Tested with images and multi-band both work.
- ImageFacet still do individual fetches (can be changed but at most it is 4 requests)
- visibleRows has been added to every b-table (visibleRows contains exactly what is on the current page and includes the sorting order)
- bulk fetching supports changing the b-table order

### DIFF
--- a/api/routes/image_pack.go
+++ b/api/routes/image_pack.go
@@ -1,0 +1,144 @@
+//
+//   Copyright © 2020 Uncharted Software Inc.
+//
+//   Licensed under the Apache License, Version 2.0 (the "License");
+//   you may not use this file except in compliance with the License.
+//   You may obtain a copy of the License at
+//
+//       http://www.apache.org/licenses/LICENSE-2.0
+//
+//   Unless required by applicable law or agreed to in writing, software
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+
+package routes
+
+import (
+	"encoding/json"
+	"io/ioutil"
+	"net/http"
+	"path"
+
+	"github.com/pkg/errors"
+	"github.com/uncharted-distil/distil-compute/metadata"
+	"github.com/uncharted-distil/distil-compute/model"
+	"github.com/uncharted-distil/distil-compute/primitive/compute"
+	"github.com/uncharted-distil/distil/api/env"
+	api "github.com/uncharted-distil/distil/api/model"
+	"github.com/uncharted-distil/distil/api/util"
+)
+
+// MultiBandPackRequest is the expected post struct for MultiBandImagePackHandler
+type MultiBandPackRequest struct {
+	Dataset string 				`json:"dataset"`
+	BandCombination string		`json:"band"`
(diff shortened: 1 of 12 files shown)


api/routes/image_pack.go
```diff
@@ -0,0 +1,197 @@
...
+		if err != nil {
+			continue
+		}
+
+		res, err := storage.FetchDataset(imagePackRequest.Dataset, false, false, false)
+		if err != nil {
+			continue
+		}
+
+		sourcePath := env.ResolvePath(res.Source, res.Folder)
```
phorne-uncharted, line 119: the storage, fetch and source path are common across all images and should be done outside of the loop

api/routes/image_pack.go
```diff
@@ -0,0 +1,197 @@
...
+				sourcePath = model.GetResourcePathFromFolder(sourcePath, dr)
+				break
+			}
+		}
+		options := util.Options{Gain: 2.5, Gamma: 2.2, GainL: 1.0, Scale: 0} // default options for color correction
+
+		imageScale := util.ImageScale{Width: ThumbnailDimensions, Height: ThumbnailDimensions}
+
+		// need to get the band -> filename from the data
+		bandMapping, err := getBandMapping(res, imageID, dataStorage)
```
phorne-uncharted, line 178: I think everything up to here is common across all images and should be done outside of the loop

api/routes/image_pack.go
```diff
@@ -0,0 +1,197 @@
...
+		funcPointer := getImages
+		if params.Band != "" {
+			// if band is not empty then get multiBandImages
+			funcPointer = getMultiBandImages
+		}
+		// channel for threads to communicate
+		result := make(chan chanStruct)
+		// ImageThreadPool is an environment variable defaults to 2 (works great with 6)
+		for i := 0; i < config.ImageThreadPool; i++ {
+			go funcPointer(params, i, config.ImageThreadPool, result, ctor, dataCtor)
```
phorne-uncharted, line 83: It may be better to have the function accept an input channel. The go routine would then listen on that channel for messages (image ids?) and would process them, writing the result to the output channel. The main process would start the go routines, write the image ids (could be done in batches) to the input channel and listen to the output channel for results.

There's also the error handling dimension that needs minor changes. They appear to be generally ignored in the two functions. I am not too sure how we want the endpoint to behave if errors occur (fail the whole batch? return only successes? return all ids with empty results for errors?), but at the very least errors should be counted and a single log entry written if any error occurs (logging the amount of errors and the last error encountered or something like that).

YohannParis (approved):
Looks good to me on the front-end, and it' so much more smoother.

PR #2432, opened 2021-03-30
update

Zac-hills:
- Added support for using an existing clone dataset (now that we support user defined column names)
- Removed onRow image fetch
- Added check to the b-table to check if the data has changed relative to the d3mindices. If the data has changed fetch the images, if not dont fetch.


YohannParis (approved):

cdbethune (approved):
will merge once conflicts resolved

PR #2457, opened 2021-04-07
Map optimization

### LINKED ISSUE #2429: Freeze when transitioning from points to quads
When working with a 30K tile dataset, there is a 4-5 second pause on my machine when transitioning from points to quads as part of a map zoom.  Profiling (see below) shows it is 100% time spent in the `areaToQuads` function of the geoplot both in processing and in garbage collecting.  Given that colour is the only thing that changes, it seems like we should be able to create the quads once, and then just switch between point and quad display as part of the zoom.  Changes to colour should just be a matter of updating the quad buffer's colour elements appropriately.

image

### LINKED ISSUE #2430: Selection slow in map drilldown view
There is a  noticeable lag  from click to the display of highlight visuals in the map drilldown view when working with large datasets.  Profiling shows that it is time entirely spent in the `areaToPoints` call.

image

Zac-hills:
closes #2429
closes #2430

- fixed some issues surrounding flexbox and the map growing
- fixed issue with no scroll bar in prediction screen

### DIFF
--- a/public/components/GeoPlot.vue
+++ b/public/components/GeoPlot.vue
@@ -816,7 +816,7 @@ export default Vue.extend({
         this.currentState.onHover
       );
     },
-    createMapLayers() {
+    createMapLayers(createOverlay = true) {
       // WebGL CARTO Image Layer
       this.tileRenderer = new lumo.TileLayer({
         renderer: new lumo.ImageTileRenderer(),
@@ -828,11 +828,13 @@ export default Vue.extend({
         lumo.loadImage(url, done); // load the image to the map
       };
       this.map.add(this.tileRenderer);
-      // Quad layer
-      this.overlay = new BatchQuadOverlay();
-      this.renderer = new BatchQuadOverlayRenderer();
-      this.overlay.setRenderer(this.renderer);
-      this.map.add(this.overlay);
+      if (createOverlay) {
+        // Quad layer
+        this.overlay = new BatchQuadOverlay();
+        this.renderer = new BatchQuadOverlayRenderer();
+        this.overlay.setRenderer(this.renderer);
+        this.map.add(this.overlay);
+      }
     },
     getInterestBounds(area: Area): LatLngBoundsLiteral {
       const xDistance = (this.drillDownState.numCols - 1) / 2;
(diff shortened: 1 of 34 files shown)


cdbethune (commented):
A few things found in testing:

- Setting the sample number via the dropdown now highlights - it was previously used to limit the number of samples fetched/rendered.

image

- Selecting via the map drilldown view is still really laggy.  It likely needs another round of optimization, but we can do that in a follow on PR.

- The map always displays with the entire world in view.  It should be zoomed to the data extends when it is first displayed.

public/components/labelingComponents/LabelGeoplot.vue
```diff
@@ -116,6 +117,13 @@ export default Vue.extend({
...
+    baselineItems(): TableRow[] {
+      const bItems =
+        datasetGetters.getBaselineIncludeTableDataItems(this.$store) ?? [];
+      return bItems.sort((a, b) => {
+        return a.d3mIndex - b.d3mIndex;
+      });
+    },
     confidenceGetter(): Function {
       if (!this.hasConfidence) {
         return (item: TableRow, idx: number) => {
```
cdbethune, line 129: item,idx can be removed

public/components/labelingComponents/LabelingDataSlot.vue
```diff
@@ -164,11 +164,7 @@ export default Vue.extend({
       return !orderBy ? false : orderBy.includes(this.labelScoreName);
     },
     dataItems(): TableRow[] {
-      const items =
-        this.viewTypeModel === GEO_VIEW
-          ? getAllDataItems(true)
-          : datasetGetters.getIncludedTableDataItems(this.$store);
-      return items;
+      return datasetGetters.getIncludedTableDataItems(this.$store);
```
cdbethune, line 167: getAllDataItems can be removed from import

public/components/GeoPlot.vue
```diff
@@ -1086,6 +1120,9 @@ export default Vue.extend({
       if (this.getCoordinateType === CoordinateType.PointBased) {
```
cdbethune, line 1124: should tableData be []TableRow?

public/util/rendering/coordinates.ts
```diff
@@ -163,3 +166,54 @@ export class PointInfo extends CoordinateInfo {
     return false;
   }
 }
+
+export function updateVertexPrimitiveColor(
+  vertices: VertexPrimitive[],
+  data: TableRow[],
+  colorFunc: (d, idx) => string,
```
cdbethune, line 173: Apply types to d, idx if possible

cdbethune (changes requested):

cdbethune (changes requested):
Found another general issue - if I click on the button to toggle the map from street map to satellite imagery all of the map content disappears.

cdbethune (approved):

PR #2458, opened 2021-04-07
Update outlier display

### LINKED ISSUE #2240: Manage how to display `metadata` fields on the front-end
As of know the _outliers detection_ results are saved as a `metadata` field in ElasticSearch, but they are not visible on the front-end. This is by design to hide some fields like _clustering_.

A better solution should be investigated.

This issue has been showcased on #2192

### LINKED ISSUE #2421: Outlier detection facet updates
The outlier detection facet should display the same icon that is used mark a column in the table view as being machine generated.  We should also not allow  the outlier detection facet to be added to the training set, as the values cannot be incorporated into a model without also adding the ability to run anomaly detection as a step in the pipeline.

YohannParis:
fix #2240 fix #2421 

- Add `distilRole` to the variable summary route.
- Update the Outlier Detection routes to behave in two steps, one to run it, the second one to save the result as a variable.
- Update the facets to not display any action for the Augmented distil role, and add an icon on the variable title.

Screen Shot 2021-04-07 at 18 04 30


YohannParis:
I have an issue because I can't create a model once the _Outlier_ variable has been added.
Which does not make sense has I made sure that the `VarDistilRoleAugmented` is not added to the solution request in the `TA2_Field` test: https://github.com/uncharted-distil/distil/blob/main/api/compute/solution_request.go#L651

PR #2464, opened 2021-04-08
Filter group variable

phorne-uncharted:
Initial switch to supporting more complicated filtering / highlighting. Only the database was updated to use the new approach. The pipeline filtering and the client were not changed. Since the client was not changed, the server assumes that all exclusive filters are part of the same filter set and all inclusive filters are part of another filter set.

The filters are now composed into filter sets which can be either inclusive or exclusive. Exclusive filter sets represent a subset of data that should be excluded from the dataset. They are processed as inclusive filter sets and then simply inverted (`NOT(equivalent inclusive filter set)`). All filter sets are evaluated and then combined via `AND` operations. Within a filter set, filters acting on the same feature (or result aspect) are combined with `OR` operations and then the clauses for the different features in the filter set are combined using an `AND` operation.

Part of this work also uncovered a lot of issues with query generation, especially with the result querying. Some data would be either ignored or duplicated when highlights were done on the result screen. There were also some issues with the highlighting of result specific fields (correctness, error, etc.) which have now been addressed.

Going forward, extending the filter set approach to pipeline filtering should be fairly straightforward. And if the client is updated to capture the idea that every click of exclude represents one distinct filter set, then it would definitely lead to cleaner code server side since all the translation between the two different structures can be done away with.

Testing was done on the baseball, auto and locust datasets. No testing was done on timeseries, and that part is definitely ripe for issues given the complicated query building that happens there. I felt it more prudent to get these changes in quicker so that we can see if we like the result.

### DIFF
--- a/api/compute/featurize.go
+++ b/api/compute/featurize.go
@@ -52,7 +52,7 @@ func (s *SolutionRequest) createPreFeaturizedPipeline(learningDataset string,
 	name := fmt.Sprintf("prefeaturized-%s-%s-%s", s.Dataset, learningDataset, uuid.String())
 	desc := fmt.Sprintf("Prefeaturized pipeline capturing user feature selection and type information. Dataset: `%s` ID: `%s`", s.Dataset, uuid.String())
 
-	expandedFilters, err := api.ExpandFilterParams(s.Dataset, s.Filters, true, metaStorage)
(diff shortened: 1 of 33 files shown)


Zac-hills (commented):
There might be away to remove the notion of include/exclude. If we move away from include/exclude to a baseline / highlight it might make it easier for the backend. The baseline is essentially the exclude but with no special parameters or cases we just treat it as a highlight but it gets applied first. I think the idea here is to remove the separation of include/exclude and just treat them all as the same set of filters just one gets applied first.

PR #2490, opened 2021-04-16
Lexbar or

### LINKED ISSUE #2474: AND/OR highlight support in Lexbar
#2464 adds support for ANDing across variables, and ORing within them.  The UI doesn't reflect this, instead showing a button for each condition:

image

This is could  be confusing for a user, as it doesn't really clarify the OR within a variable vs. the AND between.  What may be better is using a single pill per-variable, separating the ORs within a variable by comma after they have been submitted.  When the lexbar is populated as a result of facet interaction, the single variable pill is updated.  When it is populated by typing in the lexbar itself, it would use the interface for creating a new pill, but would add it back to the single variable instance on submit.

Zac-hills:
closes #2474
Peek 2021-04-16 15-21

- Lexbar now ORs duplicate highlight and filters (Note: highlight and filter pills are always separate)
- Adds & after every pill except last
- User can change OR value and it will parse out to the correct filter/highlight
- Added some bold font-weight for the operators to provide some contrast

### DIFF
--- a/public/util/lex.ts
+++ b/public/util/lex.ts
@@ -21,11 +21,11 @@ import {
   Lex,
   NumericEntryState,
   RelationState,
+  StateTemplate,
   TextEntryState,
   TransitionFactory,
   ValueState,
   ValueStateValue,
-  StateTemplate,
 } from "@uncharted.software/lex";
 import { Highlight, Variable } from "../store/dataset";
 import { Dictionary } from "./dict";
@@ -154,7 +154,7 @@ export function distilCategoryEntryBuilder(
   categorySuggestions.forEach((suggestion) => {
     const labelSuggestions =
       catVarLexSuggestions[suggestion.meta.variable.key] ?? [];
-    const branch = Lex.from("value_0", ValueState, {
+    let branch = Lex.from("value_0", ValueState, {
       allowUnknown: false,
       icon: "",
       name: "Type for suggestions",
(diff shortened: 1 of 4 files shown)


phorne-uncharted (approved):

public/util/lex.ts
```diff
@@ -91,62 +107,203 @@ class DistilRelationState extends RelationState {
...
+    ...distilNumericalEntryBuilder(allSuggestions),
+    ...distilDateTimeEntryBuilder(allSuggestions),
+    ...distilGeoBoundsEntryBuilder(allSuggestions)
   );
 }
+export function distilCategoryEntryBuilder(
+  suggestions: ValueStateValue[],
+  catVarLexSuggestions: Dictionary<unknown[]>
+): StateTemplate[] {
+  const categoryEntries = [];
```
cdbethune, line 153: As declared, this variable is effectively untyped, and will be treated as an `any[]` wherever it is used.  If the type is knowable, it should be declared as `const categroyEntries = [] as Foo[]`, which will allow the type checker to to deal with it appropriately.  Same applies to declarations of `{}`.  There are a few similar instances in this file that should be fixed if the type is known.

public/util/lex.ts
```diff
@@ -197,79 +379,120 @@ export function filterParamsToLexQuery(
...
+      if (filterMap.has(el.key)) {
+        filterMap.get(el.key).push(el);
+        return;
+      }
+      filterMap.set(el.key, [el]);
+    }
+  });
+  return { activeVariables, highlightMap, filterMap };
+}
+export function filterParamsToLexQuery(templateInfo: TemplateInfo) {
```
cdbethune, line 443: return type missing?

Zac-hills: This function directly interfaces with lex.js and is a builder for the query. The return type is dynamic and as far as I am aware it is not possible to type. The main issue is the properties scale. For example, if there are 3 highlights with the same variable type there will be 3 max and min properties (min_0, min_1, min_2, max_0, max_1, max_2).

PR #2574, opened 2021-05-07
Color scales

Zac-hills:
Peek 2021-05-07 16-21
initial color scales for variable facets

### DIFF
--- a/public/util/color.ts
+++ b/public/util/color.ts
@@ -15,6 +15,99 @@
  *    limitations under the License.
  */
 
+import { VariableSummary } from "../store/dataset";
+import { DATE_TIME_TYPE, isCategoricalType } from "./types";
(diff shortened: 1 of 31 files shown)


public/util/color.ts
```diff
@@ -15,6 +15,99 @@
...
+    key = variable.key.split(":")[1];
+  }
+  if (isCategoricalType(variable.type)) {
+    const keyMap = new Map(
+      variable.baseline.buckets.map((b, i) => {
+        return [b.key, i];
+      })
+    );
+    return (item: TableRow, idx: number) => {
+      return keyMap.get(item[key]?.value ?? item[variable.key]?.value);
```
phorne-uncharted, line 88: when would `item[key]?` not work but `item[variable.key]?` work?

PR #2635, opened 2021-05-21
Aligns prediction dataset with model dataset

cdbethune:
Adds functions to prediction dataset creation to:

1. Determine if the prediction dataset is missing variables used by the model
2. Determine if the prediction dataset has variables that don't match the order of the variables used by the model
3. Update the prediction dataset based on the findings of 1. and 2. 

This needs follow on work to address proper handling of pre-featurized datasets.

### DIFF
--- a/api/serialization/storage.go
+++ b/api/serialization/storage.go
@@ -82,31 +82,6 @@ func (d *RawDataset) AddField(variable *model.Variable) error {
 	return nil
 }
 
-// InsertField inserts a field in the dataset at the specified index, updating both the data
-// and the metadata.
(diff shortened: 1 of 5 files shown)


api/serialization/storage.go
```diff
@@ -82,6 +82,31 @@ func (d *RawDataset) AddField(variable *model.Variable) error {
 	return nil
 }
 
+// InsertField inserts a field in the dataset at the specified index, updating both the data
+// and the metadata.
+func (d *RawDataset) InsertField(variable *model.Variable, index int) error {
```
phorne-uncharted, line 87: We may want to make this function increment the index for all variables at index+1...len(variables)

PR #2691, opened 2021-06-03
Error facet fix

### LINKED ISSUE #2687: correctness filter not behaving correctly.
- When highlight both it seems to exclude the incorrect labels
- Also the facet is missing the baseline bars
Peek 2021-06-03 14-18

Zac-hills:
closes #2687
Peek 2021-06-03 15-52

### DIFF
--- a/api/model/storage/postgres/filter.go
+++ b/api/model/storage/postgres/filter.go
@@ -439,16 +439,14 @@ func (s *Storage) buildCorrectnessResultWhere(wheres []string, params []interfac
 		for _, category := range f.Categories {
 			if strings.EqualFold(category, CorrectCategory) {
 				op = "="
-				break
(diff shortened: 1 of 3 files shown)


api/model/storage/postgres/filter.go
```diff
@@ -439,16 +439,14 @@ func (s *Storage) buildCorrectnessResultWhere(wheres []string, params []interfac
 		for _, category := range f.Categories {
 			if strings.EqualFold(category, CorrectCategory) {
 				op = "="
-				break
 			} else if strings.EqualFold(category, IncorrectCategory) {
 				op = "!="
-				break
 			}
+			if op == "" {
```
phorne-uncharted, line 445: Since this check was moved inside the loop, `op := ""` should also be inside the loop.

PR #2759, opened 2021-07-08
State wrapper

Zac-hills:


api/model/filter.go
```diff
@@ -546,7 +535,7 @@ func ParseFilterParamsFromJSON(params map[string]interface{}) (*FilterParams, er
 			modeHighlights = FilterModeInverse[modeHighlights]
 		}
 		highlightSet.Mode = modeHighlights
-		filterParams.Filters = append(filterParams.Filters, &highlightSet)
+		filterParams.Highlights = append(filterParams.Filters, &highlightSet)
```
phorne-uncharted, line 538: so is everything a highlight at this point?

PR #3068, opened 2021-11-04
img and geo variables now have priority

### LINKED ISSUE #3064: Image and geobounds column should have priority for display
We normally select the top N features for display based on their PCA ranking.  When joining crop data with satellite image data, the geobounds + imagery are not in the top N features, and so end up in a undisplayed state by default.  If present, the should be given higher priority.

Zac-hills:
closes #3064

### DIFF
--- a/public/util/data.ts
+++ b/public/util/data.ts
@@ -885,8 +885,8 @@ export function sortVariablesByImportance(variables: Variable[]): Variable[] {
   });
   // give image and coordinate data priority
   return [
-    ...variables.filter(findImageAndGeoTypes),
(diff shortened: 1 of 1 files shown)


public/util/data.ts
```diff
@@ -864,16 +864,30 @@ export function sortVariablesByImportance(variables: Variable[]): Variable[] {
       rankMap = resultsGetters.getFeatureImportanceRanking(store)[solutionId];
     }
   }
+  const findImageAndGeoTypes = (v: Variable) => {
```
mj3cheun, line 867: maybe this fn should be renamed `isImageOrGeoType` or something along those lines?

Issue #3076, opened 2021-11-05
Add a way to change the number of clusters for k means in the status panel UI


Zac-hills:
![image](https://user-images.githubusercontent.com/25306965/140553666-e4acc5a4-3061-4ec2-9ecb-6ea399094b5f.png)

- There has been a recent addition to the backend that allows the clustering route to specify how many clusters
- We need some way for the user to manipulate that parameter

https://github.com/uncharted-distil/distil/blob/main/api/routes/clustering.go#L63

PR #3081, opened 2021-11-10
Add a way to change the number of clusters in the status panel

jonmarcello:
Fixes: #3076


Zac-hills (approved):

Zac-hills (changes requested):
Add some linkage between the field, fetchClustering and the clustering route.

Zac-hills (approved):

